PR #9, opened 2018-10-29
Fix seekPlayer, setVolume, setSubscriptionDuration throw exception on…

justsoft:
… Android


dooboolab:
@justsoft Wow. I can't believe I've missed this. Thanks.

PR #24, opened 2018-12-23
merged improved 662d212 from https://github.com/Shigawire

MagicalTux:
This allows setting sampleRate & numChannels for iOS (no effect on android, I don't know the apis there) without breaking compatibility with previous versions (parameters are optional, and default value keep the same behavior).

Also changes quality to max.

### DIFF
--- a/lib/flutter_sound.dart
+++ b/lib/flutter_sound.dart
@@ -83,11 +83,13 @@ class FlutterSound {
     }
   }
 
-  Future<String> startRecorder(String uri) async {
+  Future<String> startRecorder(String uri, {int sampleRate = 44100, int numChannels = 2}) async {
     try {
       String result =
           await _channel.invokeMethod('startRecorder', <String, dynamic>{
         'path': uri,
+        'sampleRate': sampleRate,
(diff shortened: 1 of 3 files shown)


lib/flutter_sound.dart
```diff
@@ -83,11 +83,13 @@ class FlutterSound {
     }
   }
 
-  Future<String> startRecorder(String uri) async {
+  Future<String> startRecorder(String uri, {int sampleRate = 44100, int numChannels = 2}) async {
```
hyochan: Could you make another `method` something like `startRecorderIOS` in order to show that it is only `iOS` compatible?

hyochan:
@MagicalTux Thank you for the `PR`. I've reviewed this and hope you to make this in `different` method until `android` side have been implemented. I'll look up on this one for android after this is merged.

lib/flutter_sound.dart
hyochan: The `PR` has been given on `android` with same feature in #29. I'll resolve this for now.

Issue #31, opened 2019-01-22
Bug on Android when startRecorder called


haideraltahan:
### Version of flutter_sound
1.2.4
## flutter doctor

>Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, v1.0.0, on Mac OS X 10.14.1 18B75, locale en-CA)
[✓] Android toolchain - develop for Android devices (Android SDK 28.0.3)
[✓] iOS toolchain - develop for iOS devices (Xcode 10.1)
[✓] Android Studio (version 3.2)
[✓] Connected device (1 available)
• No issues found!

### Platforms you faced the error (IOS or Android or both?)
This Error accurs on Android.
### Expected behavior
Expected behavior is it would start recording audio
### Actual behavior
Error Shows with this:

> E/MethodChannel#flutter_sound(13084): Failed to handle method call
E/MethodChannel#flutter_sound(13084): java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.Integer
E/MethodChannel#flutter_sound(13084): 	at com.dooboolab.fluttersound.FlutterSoundPlugin.onMethodCall(FlutterSoundPlugin.java:62)
E/MethodChannel#flutter_sound(13084): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:200)
E/MethodChannel#flutter_sound(13084): 	at io.flutter.view.FlutterNativeView.handlePlatformMessage(FlutterNativeView.java:163)
E/MethodChannel#flutter_sound(13084): 	at android.os.MessageQueue.nativePollOnce(Native Method)
E/MethodChannel#flutter_sound(13084): 	at android.os.MessageQueue.next(MessageQueue.java:326)
E/MethodChannel#flutter_sound(13084): 	at android.os.Looper.loop(Looper.java:160)
E/MethodChannel#flutter_sound(13084): 	at android.app.ActivityThread.main(ActivityThread.java:6669)
E/MethodChannel#flutter_sound(13084): 	at java.lang.reflect.Method.invoke(Native Method)
E/MethodChannel#flutter_sound(13084): 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
E/MethodChannel#flutter_sound(13084): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
[...]

PR #35, opened 2019-01-23
Fixed Android Bug

haideraltahan:
Fixing https://github.com/dooboolab/flutter_sound/issues/31

### DIFF
--- a/lib/flutter_sound.dart
+++ b/lib/flutter_sound.dart
@@ -83,7 +83,7 @@ class FlutterSound {
     }
   }
 
-  Future<String> startRecorder(String uri, {double sampleRate = 44100.0, int numChannels = 2}) async {
+  Future<String> startRecorder(String uri, {int sampleRate = 44100, int numChannels = 2}) async {
     try {
       String result =
           await _channel.invokeMethod('startRecorder', <String, dynamic>{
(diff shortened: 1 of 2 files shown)


lib/flutter_sound.dart
```diff
@@ -83,7 +83,7 @@ class FlutterSound {
     }
   }
 
-  Future<String> startRecorder(String uri, {double sampleRate = 44100.0, int numChannels = 2}) async {
+  Future<String> startRecorder(String uri, {int sampleRate = 44100, int numChannels = 2}) async {
```
hyochan: I thinks this should be `platform` specific value because in `ios`, it is receiving `double`. See the [line](https://github.com/dooboolab/flutter_sound/blob/388737aaa6451eaad98044a3a91481b29b92921c/ios/Classes/FlutterSoundPlugin.m#L147).

haideraltahan:
I have made it so that on the Android side it converts the double to int by rounding the sampling rate.

hyochan:
@haideraltahan Thank you for the contribution and updates. It's almost there. Please check one more thing I've commented.

ukrx-xamd:
Hey there hyochan - FYI :

Has this merge actually gone into 1.3.0 ? Its still broken in 1.30 

The fix works in 1.2.5, 1.2.6, 1.2.7

There is no 1.2.8 or 1.2.9 and we jump to 1.3.0 where its still broken.

Issue #11, opened 2018-11-05
Determine the current recording ampllitude


smbody:
Need the ability to determine the current recording level

ios - peakpowerforchannel
android - https://goo.gl/teko1u

smbody:
another way for android
http://qaru.site/questions/310010/android-getting-sound-frequencies-real-time

acidjazz:
@hyochan think this is possible? i could really use it!

PR #36, opened 2019-01-24
Feature/db meter

JoniDS:
Initial implementation of recording sound level (https://github.com/dooboolab/flutter_sound/issues/11)
Not very sure about the Android way of getting values, leaving it as it is until I (or someone else) finds a better way.

Note: Peak level is more resource intensive on iOS than it is on Android, so you you're targeting both platforms, don't go below 0.5 for **setDbPeakLevelUpdate** or it might become very laggy and events might get delayed (for iOS).


hyochan:
@JoniDS Thank you so much for the contribution. Could you also provide `usage guide` in the `readme` for this feature? I am not quite sure when to use this.

hyochan:
Looks great @JoniDS. Thank you for your awesome work. I've tested this and checked that this isn't breaking previous ones in both devices(ios, android). 🙌

PR #38, opened 2019-01-25
Slider widget with seek

mikeyyg96:
Before the seek would add the current millis to the new millis and not function properly. Commenting out the line that affects that works fine.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -24,6 +24,10 @@ class _MyAppState extends State<MyApp> {
   String _recorderTxt = '00:00:00';
   String _playerTxt = '00:00:00';
 
+  double slider_current_position = 0.0;
+  double max_duration = 0.0;
+
+
   @override
   void initState() {
     super.initState();
@@ -79,6 +83,9 @@ class _MyAppState extends State<MyApp> {
     try {
       _playerSubscription = flutterSound.onPlayerStateChanged.listen((e) {
         if (e != null) {
+          slider_current_position = e.currentPosition;
+          max_duration = e.duration;
+
           DateTime date = new DateTime.fromMillisecondsSinceEpoch(e.currentPosition.toInt());
           String txt = DateFormat('mm:ss:SS', 'en_US').format(date);
           this.setState(() {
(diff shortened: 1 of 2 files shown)


mikeyyg96:
You might need to change the dimensions of the container (the width most likely). I just copied the dimensions from the previous containers and put in a Slider in the child property. I am not near my personal laptop (I'm at work right now) so I can't visually see what's showing, but I'm pretty sure it functions properly; I tested it yesterday at home.

example/lib/main.dart
```diff
@@ -246,6 +253,21 @@ class _MyAppState extends State<MyApp> {
               mainAxisAlignment: MainAxisAlignment.center,
               crossAxisAlignment: CrossAxisAlignment.center,
             ),
+            Container(
+              width: 56.0,
```
mikeyyg96: Probably delete this so it can fill the dimensions of your device. Add margins to make it look nice.

hyochan:
Looks great! Approved this `PR`.

example/lib/main.dart
```diff
@@ -246,6 +253,21 @@ class _MyAppState extends State<MyApp> {
...
+              height: 56.0,
+              child: Slider(
+                value: slider_current_position,
+                min: 0.0,
+                max: max_duration,
+                onChanged: (double value) {
+                  setState(() async{
+                    String result = await flutterSound.seekToPlayer(value);
```
hyochan: Actually, you should've cast `double` to `int` conversion here since `seekToPlayer` receive `int` parameter.

example/lib/main.dart
```diff
@@ -24,6 +24,10 @@ class _MyAppState extends State<MyApp> {
   String _recorderTxt = '00:00:00';
   String _playerTxt = '00:00:00';
 
+  double slider_current_position = 0.0;
+  double max_duration = 0.0;
```
hyochan: The initial duration should be greater than `0` so I stated to `1.0` for initial state.

example/lib/main.dart
```diff
@@ -246,6 +253,21 @@ class _MyAppState extends State<MyApp> {
...
             ),
+            Container(
+              width: 56.0,
+              height: 56.0,
+              child: Slider(
+                value: slider_current_position,
+                min: 0.0,
+                max: max_duration,
+                onChanged: (double value) {
+                  setState(() async{
```
hyochan: I do not understand why you need this. This will produce tangled lifecycle.

hyochan:
I've just made update myself since I've so quickly merged yours. Checkout here #39.

PR #40, opened 2019-01-26
working_seeker_with_audio_example

mikeyyg96:
I added an example audio source that comes from my Google Firebase Storage so we can see a working demonstration to test the seeker and other player controls. The seeker works fine, you're able to press or drag anywhere in the seeker to go to that song's position. The seeker now fills the container's width, too. The onChanged property takes in a double value, which is the new value that the user has pressed or dragged to, and we use that value to seek the player to that new value, and that is why we have to convert it to an int, because flutterSound.seekToPlayer method takes an int value to seek.

### DIFF
--- a/example/test/unit_test.dart
+++ b/example/test/unit_test.dart
@@ -1,8 +1,8 @@
-import 'package:test/test.dart';
+// import 'package:test/test.dart';
 
-void main() {
-  test('my first unit test', () {
-    var answer = 11;
-    expect(answer, 11);
-  });
-}
+// void main() {
+//   test('my first unit test', () {
+//     var answer = 11;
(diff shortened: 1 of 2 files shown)


example/test/unit_test.dart
```diff
@@ -1,8 +1,8 @@
-import 'package:test/test.dart';
+// import 'package:test/test.dart';
```
hyochan: I think you may remove `unit_test.dart` from commit, if you are not making any update this this.

hyochan:
Thanks for the `PR`. Everything looks good except additional update to test file.

hyochan:
Thank you for your contribution 🤜

PR #41, opened 2019-01-27
Fix file URI for recording and playing in iOS

edman:
Android startPlayer/startRecorder assumes the given URI is absolute, while iOS assumes the URI is relative to the temporary directory. In practice this means that a valid full path given to Android would not work on iOS. This changes iOS to interpret given URIs as full paths.

E.g. passing the URI

```
/<path_to_app>/Documents/recordings/file.m4a
```

Gets translated to the (invalid, notice the double slash `"//"` and the repeated `<path_to_app>`) URI in iOS

```
/<path_to_app>/tmp//<path_to_app>/Documents/recordings/file.m4a
```

### DIFF
--- a/ios/Classes/FlutterSoundPlugin.m
+++ b/ios/Classes/FlutterSoundPlugin.m
@@ -140,8 +140,9 @@ FlutterMethodChannel* _channel;
   if ([path class] == [NSNull class]) {
     audioFileURL = [NSURL fileURLWithPath:[NSTemporaryDirectory() stringByAppendingString:@"sound.m4a"]];
   } else {
-    audioFileURL = [NSURL fileURLWithPath:[NSTemporaryDirectory() stringByAppendingString:path]];
+    audioFileURL = [NSURL fileURLWithPath:path];
   }
+  NSLog(@"HERE");
 
   NSDictionary *audioSettings = [NSDictionary dictionaryWithObjectsAndKeys:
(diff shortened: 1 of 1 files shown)


ios/Classes/FlutterSoundPlugin.m
```diff
@@ -181,12 +182,12 @@ - (void)stopRecorder:(FlutterResult)result {
 
 - (void)startPlayer:(NSString*)path result: (FlutterResult)result {
   if ([path class] == [NSNull class]) {
-    path = @"sound.m4a";
+    audioFileURL = [NSURL fileURLWithPath:[NSTemporaryDirectory() stringByAppendingString:@"sound.m4a"]];
```
hyochan: So only append temporary dir when path is not provided?

ios/Classes/FlutterSoundPlugin.m
```diff
@@ -181,12 +182,12 @@ - (void)stopRecorder:(FlutterResult)result {
...
+  } else {
+    audioFileURL = [NSURL fileURLWithPath:path];
   }
 
-  if ([[path substringToIndex:4] isEqualToString:@"http"]) {
-    audioFileURL = [NSURL URLWithString:path];
```
hyochan: Nice deletion. This might have caused some side effect earlier.

hyochan:
Looks great. I'll just merge this. Thank you 🙌.
I've sent an invitation to our team.

PR #47, opened 2019-02-14
Showing wrong recorder time text

Ajmal-M-A:
Showing wrong recorder time text when device timezone is different ( In minutes).

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:intl/intl.dart' show DateFormat;
+import 'package:intl/date_symbol_data_local.dart';
 
 import 'dart:io';
 import 'dart:async';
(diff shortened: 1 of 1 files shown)


example/lib/main.dart
```diff
@@ -45,8 +46,11 @@ class _MyAppState extends State<MyApp> {
       print('startRecorder: $path');
 
       _recorderSubscription = flutterSound.onRecorderStateChanged.listen((e) {
-        DateTime date = new DateTime.fromMillisecondsSinceEpoch(e.currentPosition.toInt());
-        String txt = DateFormat('mm:ss:SS', 'en_US').format(date);
+        initializeDateFormatting();
```
hyochan: Does this should be called in every callback? What if just initializing this once?

example/lib/main.dart
```diff
@@ -45,8 +46,11 @@ class _MyAppState extends State<MyApp> {
       print('startRecorder: $path');
```
hyochan: Hope you can remove this.

example/lib/main.dart
Ajmal-M-A: Hi Hyo Chan, Not required in every callback. Moved the initialization statement to init.

example/lib/main.dart
Ajmal-M-A: This line already exist in the source, Hopes it will be helpful for the debug purpose. So not removed.

PR #54, opened 2019-03-04
Allow BitRate customisation

JoniDS:
This PR implements bitRate customisation for both Android/iOS targets. This is particularly useful for Android where the default bitrate is very low (poor quality).
I also renamed the default android extension from .**mp4** to .**m4a** since this is the recommended for MPEG-4 audio files and it now matches iOS.


WillGeller:
I'm trying to get better quality audio coming of my android device. I set `sampleRate: 96000` what should I set `bitRate` to for higher audio quality?

PR #59, opened 2019-03-11
Fix: bitRate nullability check for iOS was wrong and throwing an exception when bitRate was not specified

JoniDS:
Also adjusted the dbPeak range (to 0-160)


ios/Classes/FlutterSoundPlugin.m
```diff
@@ -195,7 +195,7 @@ - (void)startRecorder :(NSString*)path :(NSNumber*)numChannels :(NSNumber*)sampl
                                  [NSNumber numberWithInt: [iosQuality intValue]],AVEncoderAudioQualityKey,nil];
     
     // If bitrate is defined, the use it, otherwise use the OS default
-    if(bitRate != nil) {
+    if(![bitRate isEqual:[NSNull null]]) {
```
hyochan: :+1:

Issue #53, opened 2019-03-04
output file has already been set error


ened:
### Version of flutter_sound

1.3.3

## flutter doctor

```
$ flutter doctor -v
[✓] Flutter (Channel dev, v1.3.2, on Mac OS X 10.14.3 18D109, locale en-US)
    • Flutter version 1.3.2 at /Users/ened/dev/flutter
    • Framework revision 046f960ae2 (3 days ago), 2019-02-28 23:05:45 -0800
    • Engine revision 99f3f7a9c2
    • Dart version 2.2.1 (build 2.2.1-dev.0.0 7c70ab1817)

[✓] Android toolchain - develop for Android devices (Android SDK version 28.0.3)
    • Android SDK at /usr/local/opt/android-sdk
    • Android NDK location not configured (optional; useful for native profiling support)
    • Platform android-28, build-tools 28.0.3
    • ANDROID_HOME = /usr/local/opt/android-sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1248-b01)
    • All Android licenses accepted.

[!] iOS toolchain - develop for iOS devices (Xcode 10.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 10.1, Build version 10B61
    ✗ Verify that all connected devices have been paired with this computer in Xcode.
      If all devices have been paired, libimobiledevice and ideviceinstaller may require updating.
      To update with Brew, run:
        brew update
        brew uninstall --ignore-dependencies libimobiledevice
        brew uninstall --ignore-dependencies usbmuxd
        brew install --HEAD usbmuxd
        brew unlink usbmuxd
        brew link usbmuxd
        brew install --HEAD libimobiledevice
        brew install ideviceinstaller
    • ios-deploy 1.9.4
    • CocoaPods version 1.6.1

[✓] Android Studio (version 3.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 33.3.1
    • Dart plugin version 182.5215
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1248-b01)

[✓] VS Code (version 1.31.1)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 2.24.0

[✓] Connected device (1 available)
[...]

Issue #61, opened 2019-03-15
Start recording exceptions


lizhen:
### Version of flutter_sound

flutter_sound: ^1.3.5

## flutter doctor

Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel beta, v1.2.1, on Microsoft Windows [Version 10.0.17134.471], locale zh-CN)
[!] Android toolchain - develop for Android devices (Android SDK version 28.0.3)
    ! Some Android licenses not accepted.  To resolve this, run: flutter doctor --android-licenses
[√] Android Studio (version 3.2)
[!] Android Studio (version 3.2)
    X Flutter plugin not installed; this adds Flutter specific functionality.
    X Dart plugin not installed; this adds Dart specific functionality.
[√] IntelliJ IDEA Ultimate Edition (version 2018.3)
[√] VS Code (version 1.31.1)
[√] Connected device (1 available)

### Platforms you faced the error (IOS or Android or both?)

Android

### Actual behavior

D/FlutterSoundPlugin(20851): startRecorder
D/AudioRecordPermission(20851): AudioRecordPermission
I/AudioRecordPermission(20851): remindWithResult:false
E/MediaRecorder(20851): output file has already been set
E/FlutterSoundPlugin(20851): Exception: 
E/FlutterSoundPlugin(20851): java.lang.IllegalStateException
E/FlutterSoundPlugin(20851): 	at android.media.MediaRecorder._setOutputFile(Native Method)
E/FlutterSoundPlugin(20851): 	at android.media.MediaRecorder.setInterOutputFile(MediaRecorder.java:1006)
E/FlutterSoundPlugin(20851): 	at android.hsm.HwMediaPermWrapper.setOutputFile(HwMediaPermWrapper.java:139)
E/FlutterSoundPlugin(20851): 	at android.media.MediaRecorder.prepare(MediaRecorder.java:1026)
E/FlutterSoundPlugin(20851): 	at com.dooboolab.fluttersound.FlutterSoundPlugin.startRecorder(FlutterSoundPlugin.java:162)
E/FlutterSoundPlugin(20851): 	at com.dooboolab.fluttersound.FlutterSoundPlugin.onMethodCall(FlutterSoundPlugin.java:67)
E/FlutterSoundPlugin(20851): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:201)
[...]

PR #63, opened 2019-03-18
Android command queue for recording related tasks

ened:
This adds a single threaded command queue fo all recording commands on Android (start/stop) as this could have caused interference when users started/stopped recording very quickly.

The PR also contains a few other changes to the build environment and removes a bit of dead code.
@hyochan please let me know if you'd like these changes in a separate set of PRs.

Fixes: #53, #61

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.3.6
++ Android: Adds a single threaded command scheduler for all recording related
+  commands.
++ Switch source & target compability to Java 8
++ Bump gradle plugin version dependencies
+
 ## 1.3.+
 + Support db/meter [#41](https://github.com/dooboolab/flutter_sound/pull/41)
 + Show wrong recorder timer text [#47](https://github.com/dooboolab/flutter_sound/pull/47)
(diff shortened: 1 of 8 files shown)


ened:
@hyochan Additionally, I am in favor of removing the requestPermission calls from the plugin. You can CHECK for permission and throw an error but really should not request it. That's the task for the App above.

hyochan:
@ened Understood your concern. So you want to handle `permission` differently. That would be fine. @JoniDS Could you help on reviewing if you are in favor with this?

JoniDS:
I have no issues with removing the permission handling from the library, that being said, maybe we should consider making it **optional** (enabled by default but allow the user to disable it) before taking this step, since this is a breaking change.

ened:
This PR still contains the permission handling - the change could be done in a major version upgrade in a future PR. It’s not a key issue at this moment.

hyochan:
@ened Yes. That could be handled indifferent ‘PR’. I’ll try merging this tommorow. Thanks.

hyochan:
I reverted this `PR` in #64. I've tested this and it seems it isn't working well in `example` project.

PR #66, opened 2019-03-20
Android: Recording operations on a separate command queue.

ened:
@hyochan This is the fixed version of #63.

The error in the original PR was caused by a missing "break;" statement in the `stopRecording` method handler. Classic...


ened:
@hyochan I think the change log/pub version is out of sync now.. I’ll be updating the CL next time again.

PR #67, opened 2019-03-21
WIP: Allow change of ALL settings in Example App

ened:
This is still in WIP and did uncover quite a few bugs with the implementation for me.

For example, setting invalid Audio recording encodings will break the plugin (and any future use) + it did not return a descriptive error code back to the example (or the example did not show it).

Implementation bug fixes will be provided in separate PRs as I discover them.

Also need to decide
- if we want to filter iOS/Android settings on the appropriate platforms (I guess it's fine to show all).
- if bitrate/number of channels/sample rate settings should be dropdowns instead

@hyochan WDYT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.3.6
++ Android: Adds a single threaded command scheduler for all recording related
+  commands.
++ Switch source & target compability to Java 8
++ Bump gradle plugin version dependencies
+
 ## 1.3.+
 + Support db/meter [#41](https://github.com/dooboolab/flutter_sound/pull/41)
 + Show wrong recorder timer text [#47](https://github.com/dooboolab/flutter_sound/pull/47)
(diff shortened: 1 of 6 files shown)


hyochan:
@ened Thanks for continuous support for your `PR`. 
1. if we want to filter iOS/Android settings on the appropriate platforms (I guess it's fine to show all).
   - Agree with showing all.
2. if bitrate/number of channels/sample rate settings should be dropdowns instead
   - The dropdown shouldn't matter since this is just an example project. Do this as much as your favor. Thanks for your additional concern in `example` too.

hyochan:
@ened Is this ready to be merged?

ened:
Still WIP as I’d like to add proper error reporting (and handling in example) as well. Like said in the initial comment, setting invalid config will kill the plugin permanently currently.

hyochan:
Since this `PR` haven't been updated for a while, I'll close it!

Issue #70, opened 2019-04-04
seekToPlayer error 


zfha:
in iOS Platform
_TypeError (type 'int' is not a subtype of type 'String')

hyochan:
Can you also try `example` project?

nicholas-vella:
I've just opened a pull request for this !72

CoreyCole:
I see this is merged, but is it live in `1.4.0`? I'm getting the error on iOS about the return type of seek:
```
[VERBOSE-2:ui_dart_state.cc(148)] Unhandled Exception: type 'int' is not a subtype of type 'String'
#0      MethodChannel.invokeMethod (package:flutter/src/services/platform_channel.dart:302:13)
<asynchronous suspension>
#1      FlutterSound.seekToPlayer (package:flutter_sound/flutter_sound.dart:190:24)
<asynchronous suspension>
#2      AudioService.seekPlayer (package:mokri_storytime/src/blocs/services/audio_service.dart:219:19)
```

PR #72, opened 2019-04-09
iOS: Fixes for seekToPlayer 

nicholas-vella:
`seekToPlayer` on Android returns a String, and on iOS was returning an int. This is reflected in Issue #70. This PR fixes that, making iOS match the Android behaviour.
The iOS `AVAudioPlayer` class's `currentTime` property is measured in seconds instead of milliseconds. This PR converts milliseconds into seconds when setting the current time in the `seekToPlayer` function.
Additionally, the `seekToPlayer` function wasn't calling the `updateProgress` Dart method, which was happening on Android. This PR also adds equivalent functionality to iOS.

### DIFF
--- a/ios/Classes/FlutterSoundPlugin.m
+++ b/ios/Classes/FlutterSoundPlugin.m
@@ -357,8 +357,9 @@ NSString* status = [NSString stringWithFormat:@"{\"current_position\": \"%@\"}",
 
 - (void)seekToPlayer:(nonnull NSNumber*) time result: (FlutterResult)result {
   if (audioPlayer) {
-    audioPlayer.currentTime = [time doubleValue];
-    result(time);
+      audioPlayer.currentTime = [time doubleValue] / 1000;
+      [self updateProgress:nil];
+      result([time stringValue]);
   } else {
     result([FlutterError
         errorWithCode:@"Audio Player"


ios/Classes/FlutterSoundPlugin.m
```diff
@@ -357,8 +357,9 @@ - (void)resumePlayer:(FlutterResult)result {
 
 - (void)seekToPlayer:(nonnull NSNumber*) time result: (FlutterResult)result {
   if (audioPlayer) {
-    audioPlayer.currentTime = [time doubleValue];
-    result(time);
+      audioPlayer.currentTime = [time doubleValue] / 1000;
+      [self updateProgress:nil];
```
hyochan: Is this line needed? `[self updateProgress:nil]`?

nicholas-vella: This line isn't strictly necessary. After a bit more investigation, this can be handled by the host application's logic. However, having this line provides a confirmation that the seekToPlayer function correctly updated the AVAudioPlayer's currentTime. I'm open to removing or leaving it, either way.

hyochan: Alright! Let's keep this update for now.

PR #80, opened 2019-05-07
fix(example): remove milliseconds conversion

CoreyCole:
`seekToPlayer()` takes milliseconds as input on iOS now. We can remove this conditional check from the example because it is no longer valid.


hyochan:
Thanks for the contribution! 🎉

PR #123, opened 2019-08-16
Improved db calcs.

bsutton:
The db calcs were constantly saturating at 160db even with a low volume recording.
Also fixed some state inconsistency with method callbacks.
Improved some the exceptions been thrown.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,4 +181,5 @@ _dbPeakSubscription = flutterSound.onRecorderDbPeakChanged.listen((value) {
 ## Help Maintenance
 I've been maintaining quite many repos these days and burning out slowly. If you could help me cheer up, buying me a cup of coffee will make my life really happy and get much energy out of it.
 <br/>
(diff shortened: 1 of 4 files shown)


hyochan:
@bsutton Thank you for your contibution! Could you please `rebase`? There is a conflict with the `master` branch.

bsutton:
I think this should now be correct as github is reporting that it has no conflicts.

jam33:
What is the reasonable range for db here?  I agree that previously it was pulling 160db at very reasonable/low levels.  Now it seems the max I can register is 70db (computer speakers turned full blast right next to phone mic).  To be exact it was: 70.00061068314733.

The demo still seems to base the UI "feedback bar" (progress indicator) off of 160:
`value: 100.0 / 160.0 * (this._dbLevel ?? 1) / 100,`

So, the "green" bar never gets more than 1/2-1/3 of the way across the "red" indicator bar.

bsutton:
80db is a reasonable range.

PR #131, opened 2019-09-10
Stopped recording generating infinite db values.

bsutton:
Added check for a maxAmplitude of zero which can happen if the mic is off which was resulting in db being set to infinity. Db is now set to zero.

This problem may only be an issue with the simulator, but its annoying enough in development that this simple fix is warrented.

This pull also includes some code cleanup to remove a number of lint warnings.


android/src/main/java/com/dooboolab/fluttersound/FlutterSoundPlugin.java
```diff
@@ -199,11 +199,17 @@ public void startRecorder(int numChannels, int sampleRate, Integer bitRate, int
             // https://stackoverflow.com/questions/10655703/what-does-androids-getmaxamplitude-function-for-the-mediarecorder-actually-gi
             // 
             double ref_pressure = 51805.5336;
-            double p = maxAmplitude / ref_pressure;
+            double p = maxAmplitude  / ref_pressure;
```
hyochan: I think you can revert this.

Issue #139, opened 2019-10-04
1.4.8 Android startRecorder FileNotFoundException


Marco-Bronner:
### Version of flutter_sound 
^1.4.8

## flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, v1.9.1+hotfix.4, on Microsoft Windows [Version 10.0.18362.387], locale de-DE)
[√] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
[√] Android Studio (version 3.5)
[√] Connected device (1 available)

• No issues found!

### Platforms you faced the error (IOS or Android or both?)
Android, as I have no IOS to test

### Expected behavior
Recording starts as usual.

### Actual behavior
calling _flutterSound.startRecorder with sepcific path resulst in Exception. Calling with null as uri works.

E/FlutterSoundPlugin(24685): Exception: 
E/FlutterSoundPlugin(24685): java.io.FileNotFoundException: /storage/emulated/0/data/user/0/dt.de.voicematching/app_flutter/IfJMfNhpB7NsQmcmbtMc: open failed: ENOENT (No such file or directory)
E/FlutterSoundPlugin(24685): 	at libcore.io.IoBridge.open(IoBridge.java:496)
E/FlutterSoundPlugin(24685): 	at java.io.RandomAccessFile.<init>(RandomAccessFile.java:289)
E/FlutterSoundPlugin(24685): 	at java.io.RandomAccessFile.<init>(RandomAccessFile.java:152)
E/FlutterSoundPlugin(24685): 	at android.media.MediaRecorder.prepare(MediaRecorder.java:1046)
E/FlutterSoundPlugin(24685): 	at com.dooboolab.fluttersound.FlutterSoundPlugin.startRecorder(FlutterSoundPlugin.java:167)
E/FlutterSoundPlugin(24685): 	at com.dooboolab.fluttersound.FlutterSoundPlugin.lambda$onMethodCall$0$FlutterSoundPlugin(FlutterSoundPlugin.java:70)
E/FlutterSoundPlugin(24685): 	at com.dooboolab.fluttersound.-$$Lambda$FlutterSoundPlugin$Uw7hmMVnMhHSlTmtp8I_WAUlmZE.run(Unknown Source:8)
E/FlutterSoundPlugin(24685): 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:462)
[...]

PR #143, opened 2019-10-26
Bugfix #139

tvw:
I fixed both, the Android and IOS variant, but I only could test the Android variant.

So please check the IOS variant change since the fix was the first line of ObjectiveC I ever wrote.

The original code prefixes the given path given to ```startRecorder``` with another path, which results in a broken path.

### DIFF
--- a/android/src/main/java/com/dooboolab/fluttersound/FlutterSoundPlugin.java
+++ b/android/src/main/java/com/dooboolab/fluttersound/FlutterSoundPlugin.java
@@ -143,8 +143,6 @@ public class FlutterSoundPlugin implements MethodCallHandler, PluginRegistry.Req
 
     if (path == null) {
       path = AudioModel.DEFAULT_FILE_LOCATION;
-    } else {
-      path = Environment.getExternalStorageDirectory().getPath() + "/" + path;
     }
 
     if (this.model.getMediaRecorder() == null) {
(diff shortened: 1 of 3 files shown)


tvw:
@hyochan No I did not. I checked it with my own project in Android and it works. I did not test IOS.

But I checked the example code, and it depends on the bug I have fixed, which means, it is buggy itself.

```dart
 String path = await flutterSound.startRecorder(Platform.isIOS ? 'ios.m4a' : 'android.mp4');
```

The method call must receive a path, not just a filename. In order to make this run, you cannot achieve this only in flutter. You need another library like [path_provider](https://pub.dev/packages/path_provider).

Or you use the default file, the method will use, when you pass ```null```:

```dart
 String path = await flutterSound.startRecorder(null);
```
I have not tried this with the example but when investigating the problem with my app on Android and it worked.

tvw:
@hyochan I also updated the example.

hyochan:
@tvw Could you also give a scenario where we can use `path_provider` to give `custom` pass and the readme? Currently, it only does `startRecorder(null)` and I can't test any other things. If I pass `startRecorder('ios.m4a')` like it was working previously, it won't work.

Also, changing the `readme` for this would be wonderful.

hyochan:
@tvw Any news on this?

PR #168, opened 2019-12-10
Fixes returned filePath on iOS

phranck:
Actually it returns the absolute string of a so called file URL which starts with `file:///`. This isn't a POSIX compliant file path which always starts with a slash `/`.

The Android version of this plugin returns the result of `getPath` which gives us a POSIX compliant path, too.

@hyochan It would be very nice to have this PR in a current release of the plugin as soon as possible. Actually we are working with our fixed fork, which we would like to get rid of. Thanks a lot!

### DIFF
--- a/ios/Classes/FlutterSoundPlugin.m
+++ b/ios/Classes/FlutterSoundPlugin.m
@@ -243,7 +243,7 @@ NSString* status = [NSString stringWithFormat:@"{\"current_position\": \"%@\"}",
         [self startDbTimer];
   }
 
-  NSString *filePath = self->audioFileURL.absoluteString;
+  NSString *filePath = self->audioFileURL.path;
   result(filePath);
 }
 


Larpoux:
I had exactly the same problem with my own fork.
I fixed this bug replacing the 2 last lines of startRecorder on iOS by the following:
```
  NSString *filePath = self->audioFileURL.path; // A real file path and not a URL to the file [LARPOUX]
  result(filePath);
```

hyochan:
It sounds reasonable so I am happy to merge this. However, I have one question. Don't we need to append `file://` then? In `ios` because android is returning that without another slash?

hyochan:
@phranck It sounds reasonable so I am happy to merge this. However, I have one question. Don't we need to append `file://` then? In `ios` because android is returning that without another slash?

phranck:
> It sounds reasonable so I am happy to merge this. However, I have one question. Don't we need to append `file://` then? In `ios` because android is returning that without another slash?

@hyochan Thanks for merging! Nope, there is nothing to append. The result should return an absolute path, nothing more. And absolute path's doesn't contain a scheme. They always start with a `/`. We tested it on Android too (in Emulator with API level 28) and the [corresponding part in code](https://github.com/dooboolab/flutter_sound/blob/master/android/src/main/java/com/dooboolab/fluttersound/FlutterSoundPlugin.java#L146) returns just an absolute path, too. There is no scheme whatsoever.

PR #169, opened 2019-12-10
Create a new method : startPlayerFromBuffer

Larpoux:
Create a new method : 'startPlayerFromBuffer' to play from a memory buffer instead of a file.
Tested on iOS and Android.

+ Fix: Fixes returned filePath on iOS #168


Larpoux:
I had exactly the same problem with my fork.
I fixed this bug replacing the 2 last lines of startRecorder on iOS by the following:
```
  NSString *filePath = self->audioFileURL.path; // A real file path and not a URL to the file [LARPOUX]
  result(filePath);
```

Larpoux:
I realize today that the push I did yesterday was not good : too many files modified without a good reason. I will redo a better 'pull request' soon. Sorry.

PR #182, opened 2020-01-04
iOS record/play opus

Larpoux:
Now flutter_sound can record and play regular **ogg/opus** audio files, and not only this stupid _caf/opus_ specific to Apple. This works OK with files (i.e. 'foo.opus'), but not yet with playFromBuffer().

We do not want to use something else than the actual "AVRecorder" that flutter_sound actually uses. So the solution was to remove the stupid/Apple envelope after recording, and add a stupid/Apple envelope before playing. With that hack, Apple is happy (and me too :-D ).

Adding/removing the CAF envelope is a very lite process (we do not decode or re-encode the stream) so we will not have any noticeable delay. I use the FFmpeg library to manipulate the file format. FFmpeg is LGPL. I do not think that it will be a problem for flutter_sound.

Here is the new Codec compatibility table:
Actually, the following codecs are supported by flutter_sound:

| | AAC  | OGG/Opus  | CAF/Opus | MP3 | OGG/Vorbis | PCM |
| :------------ |:---------------:| :---------------:| :-----| :-----| :-----| :-----|
| iOS encoder| Yes | Yes | Yes | No | No | No |
| iOS decoder| Yes | Yes | Yes | Yes | No | Yes |
| Android encoder| Yes | No | No | No | No | No |
| Android decoder| Yes | Yes | No | Yes | Yes | Yes |

This table will eventually be upgrated when more codecs will be added.

Note: I do not master very well pod, podspec, flutter yaml, xcode workspaces, Frameworks, and all those environment tools. So I am not entirely confident that we will not have to fix some dependencies declaration before publishing the plugin.


Larpoux:
Note : 
- I did not push on github the 3 FFmpeg libraries that I use. You need to do `./build-ffmpeg-iOS-framework.sh` to recreate the FFmpeg libraries

- FFmpeg needs libz, libz2 and libiconv

hyochan:
@Larpoux Would you like to resolve the conflicts to master branch?

Larpoux:
Sure :-)
But I did not notice any conflits

Larpoux:
Oh! Yes, I see. There is a conflict in the README. I will fix that immediately.
PS: I am currently working on supporting “playFromBuffer()” with the OPUS_CODEC

Larpoux:
Hi Hyo. The 2 conflits are not very serious.
- Just a modification in the README to specify that we now support OPUS on iOS
- And a modification in the array of supported codec for iOS

But I do not master very well github and I do not know how to resolve those 2 conflicts. Do I must cancel my pull request and do another one ?

Larpoux:
OK Hyo, I think the conflicts are solved.

hyochan:
> Notes :
> 
> * I did not push on github the 3 FFmpeg libraries that I use. You need to do `./build-ffmpeg-iOS-framework.sh` and `build-ffmpeg.sh` to recreate the FFmpeg libraries and Framework
> * FFmpeg needs libz, libz2 and libiconv
> * Play opus asset does not work yet, because we use playFromBuffer() to play assets

Thanks for `resolving` conflict. I have a question about this. What causes problem when we do not import `FFmpeg` in our library? Would some feature you've implemented won't work?

Larpoux:
The  (https://github.com/Canardoux/flutter_sound/blob/ios-play-opus/ios/Classes/opus2caf.m)[iOS module] needs FFmpeg for replacing the CAF (Apple specific) file format after recording OPUS, by a regular/standard OGG format. In the other direction, FFmpeg is used to replace the regular/standard OGG format by CAF, so that Apple is pleased and can play it.

CAF is really stupid proprietary file format from Apple. (Apple always thinks that everything on earth is Apple, and only Apple, and that they do not need to bother with compatibility issues. This is a terrible behaviour for developers and users).

My App is based on a Korean/English dictionary, with hundred of thouthands of records in Korean and in English. Those records are in standard OGG/OPUS format, stored in a SQL database. I really need a way to play those records both with Android and iOS.

Actually, flutter_sound is based on iOS upon AVAudioRecorder and AVAudioPlayer. I like that very much, because flutter_sound is simple, without any complicate solutions with threads and other difficult solutions.
But we cannot plug any other codec to AVAudioPlayer, and we must use the codecs that Apple provides. Fortunately, Apple supports OPUS, but encapsulates it in its proprietary format.
This is the reason I use now FFmpeg.
Other solutions would be to decode OPUS with a standard codec and not using the Apple one. But if we do that, we will have to forget AVAudioPlayer, and it would not be flutter_sound anymore.
The last solution, for me, would be to re-encode my database records from OPUS to AAC. AAC is not so bad, but I would prefer to use OPUS.

Larpoux:
I really does not like the idea of maintaining two different databases, one for Android with OGG/OPUS format and another one with CAF/OPUS format. But I admit that it would be another solution if we cannot support OGG/OPUS in flutter_sound.

Larpoux:
The very last step for me, will be to understand why we cannot record OGG/OPUS with Android. When this point will be fixed, we will have OGG/OPUS totally supported by iOS and Android. But I will work on that point later : I have now much work to do on my own App, and not flutter_sound.

Larpoux:
OK Hyo, play from buffer (and play from asset) is now supported for iOS, with OGG/OPUS.
I use two temporary files : `tmp/flutter_sound.opus` and `tmp/flutter_sound.caf`.
We can probably do better (without temporary files) but I do not have enough time to investigate, and flutter_sound is OK for my own use like that.

hyochan:
I really appreciate all your hard work and make evolutionary to this plugin 💯 🥇 

> OK Hyo, play from buffer (and play from asset) is now supported for iOS, with OGG/OPUS.
> I use two temporary files : `tmp/flutter_sound.opus` and `tmp/flutter_sound.caf`.
> We can probably do better (without temporary files) but I do not have enough time to investigate, and flutter_sound is OK for my own use like that.

Does above means that we no longer need ffmepg to play `OGG/OPUS` in `ios`?

Larpoux:
Yes, **we still need  ffmepg** to play OGG/OPUS in iOS.
Maybe we can try **NOT** to link FFmpeg for users who do not need OGG/OPUS, but I do know how : the linker will complain if no FFmpeg.

Flutter_sound is great. **Simple** and great. One day, Flutter_sound can be **THE** flutter audio plugin. (I looked rapidly to others plugins but I was not really pleased : they are often too complicated. Better to relay on native tools, as flutter_sound does.)

hyochan:
@Larpoux Thanks for your promising opinion. So you think we can just merge this without `ffmpeg` installed since it only tells you to link it when dev tries to play `OGG/OPUS`. Am I right?

ios/.sf
```diff
@@ -0,0 +1,15 @@
...
+9WxsTkoaAFPCPXkv5GZ95W73p41G1zPFh5kOHwnli7Kvb6IKSQq/yrfTf5SOTUox
+85JAYk3pCPqVSA/v5i6z1e11382jdAszNA8+B916gjK84YRFAkYB40Y+BHvsrqHW
+9XL0o6u0BZmUkx+hx78IrwmIewqbq2+6lU+vOTFWevl1pYHQIm8x3IwazM275zu/
+e4Rth3sDGiELASjW+LZs2+PwchL7uyxGmlF/YuHJDePWWEl18Hk8lVBebjiY1DqA
+ZpaC9VdHpuioK5JxBcZktqw18hVR8hkVQpJUT0PokkEsEf2+oeymY6LgFbnIrO8/
+SwBgqjHHRKwbzrLok3cRUiC+gxNWjHi+x+W/QVG4l4On/bsCT/EAxiF+Q+p1U7OC
+4YkxNF7A2l1PcNk+PAhy4VyHfOtidcGeRomHj5DiS98QQJCLUaVE4kmVrF1wDYlO
+pc9qKtyvhFGz/eEpXmLV9gzHD3S7pKdBU0NAU8wVeeaGOGyX9ISpztQlvEWlPyEQ
+ke4M+Pj1CfOQxFc5rX6EoztI+qHogEUcBD9FBYtFswBAcSG9kYhIuORR44LPB6yF
+m98e+PfyXGVLGI5GFNV1Sw==
```
hyochan: Could you tell what this file does?

Larpoux: ios/.sf and ios/.travis.yml are par of [FFmpeg build script](https://github.com/kewlbear/FFmpeg-iOS-build-script). I do not know what they are used for.

Larpoux:
> @Larpoux Thanks for your promising opinion. So you think we can just merge this without `ffmpeg` installed since it only tells you to link it when dev tries to play `OGG/OPUS`. Am I right?

We can probably use a `config.h` so that our user can specify if he/she is interested by OGG/OPUS.
Something like : 
```
#define OGG_OPUS 1
```
and use conditional compilation in FlutterSoundPlugin.m

But you definitely have to provide the FFmpeg framework for the users who need OGG/OPUS (me, for example ;-)

Larpoux:
This branch is **not ready** to be published.
It needs some work and actually I do not have enough time for that.
Maybe later this year ...

PR #189, opened 2020-01-09
Added Media Controls on Lock Screen

salvatore373:
After this merge the plugin will follow the Android and iOS guidelines for media player. I added the media controls on the lock screen for when the user starts playing an audio file.

To accomplish this I had to make some changes to the original code:
- I adapted the Android media player to the Android guidelines for background services like an audio player;
- I added the `initialize` and `releaseMediaPlayer` methods;
- I changed the startPlayer method.

As I edited the Android native code to follow the platform guidelines, I created a `MediaBrowserService` that is initialized by a `MediaBrowser`. I won’t dive into details about why it is better for an application to have those classes, but they were needed to display a notification to handle the media playback in the right way.
The `MediaBrowser`, which in turn initializes the `MediaBrowserService`, must be initialized before using the audio player. I couldn’t initialize it in the `startPlayer` method: for this reason I created the `initialize` method, where I create the `MediaBrowser`. The `intialize` method does nothing in the iOS native code because the audio player is initialized directly in `startPlayer`.
The Android guidelines requires the developer to clean up the device resources from a service (like an audio player) when the user no longer needs it. The best way to do it was to create another method, the `releaseMediaPlayer` method. A similar job was previously done by the `stopPlayer`, but the new method is more complete because it deals with the `MediaBrowser` and the `MediaBrowserService` and resets them.

With this changes the application was able to display a notification with the media controls to handle the audio player playback when the player started to play a sound. But no one was providing the service with the information to show in the notification. A title, an author and an album art were needed. So, I changed the arguments of the `startPlayer`. Now, instead of a `String` with the URI that points to the audio file to play, it requires a `Track` object, that stores the title and author names, the URL to an album art and the `String` with the URI that points to the audio file to play. The `startPlayer` method also requires two booleans that are used to decide whether the “skip forward” and “skip backward” buttons in the notification should be enabled or not.
The information provided to `startPlayer` can change whenever the audio player starts playing a new track, while the actions to perform when the skip buttons are pressed should never change and should be equal for each track that gets played. For this reason the functions to execute as callbacks for the skip buttons are provided only once to the `initialize` method.

Furthermore, I added a `Stream` that allows to listen to changes of the audio player playback state (when the player starts playing, when it is paused and when it is stopped).

I include two screencast (one for Android and iOS) showing the results of my changes. You will basically see no changes, except for the notification that gets shown when the audio player starts playing.
screencast-android
screencast-ios

This is the first time that I contribute to an open source project, then I’m sorry for any error that you will encounter. And, since I’m not native language, I’m sorry for my English.
However I am open to any discussion about my code and my changes and to clarify my decisions.


salvatore373:
Could you please explain me the reasons of the conflict?

hyochan:
Your pr have been made from old commits which is behind the master.
Try reading about [forking workflow](https://link.medium.com/4PdBWTaB82)

salvatore373:
Thank you, I will try to fix it.

salvatore373:
I resolved all the conflicts and opened a new PR (#202 ). So I can close this one.

PR #209, opened 2020-01-28
Use AAC-LC format instead of MPEG-4

jonasfranz:
At the moment we use MPEG-4 as format for android recordings and AAC LC for ios recordings. This PR sets AAC LC as default for AAC recordings under android to have the same formats as under ios.

This may be breaking.

### DIFF
--- a/android/src/main/java/com/dooboolab/fluttersound/FlutterSoundPlugin.java
+++ b/android/src/main/java/com/dooboolab/fluttersound/FlutterSoundPlugin.java
@@ -111,8 +111,8 @@ public class FlutterSoundPlugin implements MethodCallHandler, PluginRegistry.Req
   };
 
   static int formatsArray[] = {
-      MediaRecorder.OutputFormat.MPEG_4 // DEFAULT
-    , MediaRecorder.OutputFormat.MPEG_4 // CODEC_AAC
+      MediaRecorder.OutputFormat.AAC_ADTS // DEFAULT
+    , MediaRecorder.OutputFormat.AAC_ADTS // CODEC_AAC
     , sdkCompat.OUTPUT_FORMAT_OGG       // CODEC_OPUS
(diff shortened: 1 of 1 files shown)


Larpoux:
I am really not a specialist with audio encoders.
But if MPEG-4 and AA c-LC are different, should we not have two different codecs declared in flutter_sound ?

Larpoux:
I did not realize that encoding are not same on iOS and Android with 't_CODEC.CODEC_AAC'. If this is really the case, I agree that we must fix that. I will look how to declare 2 codecs in flutter_sound.

Larpoux:
OK, I did some search (thank you Google). MPEG4 is a file format and not a codec.
The little problem in flutter_sound is that we have juste one enum, to specify at the same time the file-format and the codec. In the actual version, flutter_sound muxes AAC in a MPEG4 file-format on Android. This is not completly stupid, but I agree that this is not homogeneous with iOS.

I suggest to integrate now your patch in flutter_sound, and have some reflexion if we must add a new t_CODEC for ```MPEG4/AAC```, use 2 parameters in startRecorder, one for file-format and another for the codec , .... or do nothing.

Larpoux:
If you ask me, I will vote to use two parameters in startRecorder. One for the file-format and another for the codec.
I already had some mess with OPUS which is used normally in an OGG file-format, but that Apple encapsulates in its own CAF format.
Also some interesting things with OGG which is used with VORBIS codec or OPUS codec.

We certainly can maintain backward compatibility if the user does not specify the file-format that he wants.

jonasfranz:
If we introduce an a new parameter `format` it should have the same default value for both iOS and Android. I think there is no other way then breaking existing implementation without providing the same defaults for both platforms.

Larpoux:
Yes, you are both right. I agree with this PR.
For me, this can be included in next flutter_sound version.

PR #228, opened 2020-02-25
Fix path issue when playing audio on android

claptimes5:
# Flutter Doctor
```Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, v1.12.13+hotfix.8, on Mac OS X 10.15.3
    19D76, locale en-US)

[✓] Android toolchain - develop for Android devices (Android SDK
    version 29.0.2)
[✓] Xcode - develop for iOS and macOS (Xcode 11.3.1)
[✓] Android Studio (version 3.6)
[✓] VS Code (version 1.42.1)
[✓] Connected device (1 available)
```

# Android issue
Tested on a real device (Galaxy Note 9) 

## Expected behavior
Calling startPlayer() with the path of an audio file on external storage actually plays the file. 

## Actual behavior
Result is an error. After investigating futher, the `prepare()` method was causing the issue. The path desired path to the file should start with `file://`. 

Here is the flutter log for the error:

```I/flutter (19693): startPlayer called
I/flutter (19693): playing status: t_AUDIO_STATE.IS_STOPPED
I/flutter (19693): stopping player
I/flutter (19693): stopPlayer called
I/flutter (19693): playing status: t_AUDIO_STATE.IS_STOPPED
I/flutter (19693): playing file
I/flutter (19693): /data/user/0/com.claptimes.myapp/app_flutter/my_app/2020-02-25T11:50:24.165751.acc
V/MediaPlayer-JNI(19693): native_setup
V/MediaPlayerNative(19693): constructor
V/MediaPlayerNative(19693): setListener
V/MediaPlayerNative(19693): setVideoSurfaceTexture
V/MediaPlayerNative(19693): prepare
V/MediaPlayerNative(19693): message received msg=300, ext1=0, ext2=0
V/MediaPlayerNative(19693): Received SEC_MM_PLAYER_CONTEXT_AWARE
V/MediaPlayerNative(19693): callback application
V/MediaPlayerNative(19693): back from callback
V/MediaPlayerNative(19693): message received msg=100, ext1=1, ext2=-2147483648
E/MediaPlayerNative(19693): error (1, -2147483648)
V/MediaPlayerNative(19693): signal application thread
V/MediaPlayerNative(19693): prepare complete - status=1
E/FlutterSoundPlugin(19693): startPlayer() exception
I/flutter (19693): error: Exception: PlatformException(ERR_UNKNOWN, ERR_UNKNOWN, Prepare failed.: status=0x1)
I/flutter (19693): stopPlayer called
I/flutter (19693): playing status: t_AUDIO_STATE.IS_STOPPED
```

# Fix
Prefixing the path string with `file://` resolves the issues. I've included a PR which works on my device.

### DIFF
--- a/android/src/main/java/com/dooboolab/fluttersound/FlutterSoundPlugin.java
+++ b/android/src/main/java/com/dooboolab/fluttersound/FlutterSoundPlugin.java
@@ -4,6 +4,7 @@ import android.Manifest;
 import android.content.pm.PackageManager;
 import android.media.MediaPlayer;
 import android.media.MediaRecorder;
+import android.net.Uri;
 import android.os.Build;
 import android.os.Handler;
 import android.os.SystemClock;
@@ -420,7 +421,9 @@ public class FlutterSoundPlugin implements MethodCallHandler, PluginRegistry.Req
       if (path == null) {
         this.model.getMediaPlayer().setDataSource(AudioModel.DEFAULT_FILE_LOCATION);
       } else {
-        this.model.getMediaPlayer().setDataSource(path);
+        Uri uriFileName = Uri.fromFile(new File(path));
+        String uriStr = uriFileName.toString();
+        this.model.getMediaPlayer().setDataSource(uriStr);
       }
 
       this.model.getMediaPlayer().setOnPreparedListener(mp -> {
@@ -486,6 +489,8 @@ public class FlutterSoundPlugin implements MethodCallHandler, PluginRegistry.Req
     } catch (Exception e) {
       Log.e(TAG, "startPlayer() exception");
       result.error(ERR_UNKNOWN, ERR_UNKNOWN, e.getMessage());
+//      Log.e(TAG, e.getStackTrace().toString());
+//      Log.e(TAG, Thread.currentThread().getStackTrace().toString());
     }
   }
 
--- a/android/src/main/java/com/dooboolab/fluttersound/FlutterSoundPlugin.java
+++ b/android/src/main/java/com/dooboolab/fluttersound/FlutterSoundPlugin.java
@@ -489,8 +489,6 @@ public class FlutterSoundPlugin implements MethodCallHandler, PluginRegistry.Req
     } catch (Exception e) {
       Log.e(TAG, "startPlayer() exception");
       result.error(ERR_UNKNOWN, ERR_UNKNOWN, e.getMessage());
-//      Log.e(TAG, e.getStackTrace().toString());
-//      Log.e(TAG, Thread.currentThread().getStackTrace().toString());
     }
   }
 


hyochan:
We decided to use the external library [path_provider](https://pub.dev/packages/path_provider) for this. @Larpoux Could you also give a look at this? Since I remember, we talked about this earlier.

Larpoux:
@hyochan : I think this issue is different than the one we talk together about.
It was related with bad paths given to the recorder(bad access authorisation).

This issue is not about recording, but playing an existing sound file.
It is not clear for me, why a URI is better than a full path 🙄 
I will check this PR more deeply, and I will give you my position.

Larpoux:
I do not agree with this patch : 
Actually, the caller can call `startPlayer()` with a URI.
This is used, for example, by our Example Demo App to play a remote file.
If the caller wants to play `file://data/user/0/com.claptimes.myapp/app_flutter/my_app/2020-02-25T11:50:24.165751.acc`, he just have to specify this URI when calling `startPlayer()`.

This is not flutterSound responsibility to hack the parameter given by the caller.

Finally, I do not see any reason which explain why a local file could be accessed when a URI is provided, but not when a complete file path is provided.
Note : Probably it would be possible to introduce a new API verb : `startPlayerFromUri()` if `startPlayer()` will not enough. But I am not sure that actually this is necessary, and we will need to check what to do on iOS.

claptimes5:
@Larpoux Thanks for checking into this. The issue may be caused by where I was storing the files. I'll check later and post here, but I was previously using `getApplicationDocumentsDirectory()` when storing my files and just switched to `getExternalStorageDirectory()` on Android.

claptimes5:
> @Larpoux Thanks for checking into this. The issue may be caused by where I was storing the files. I'll check later and post here, but I was previously using `getApplicationDocumentsDirectory()` when storing my files and just switched to `getExternalStorageDirectory()` on Android.

This guess was not correct. The files do not play and give the same error. 

@Larpoux The only reason I'm submitting this as a patch instead of just passing through the URI, is that calling `startPlayer(absolute_path)` does not work for me and should work according to the documentation.

Larpoux:
OK @claptimes5 . I will look to that, again.
Are you sure that it is OK to have ':' in your path name ? I really think it could be a problem.

Maybe we will need two API verbs : 
- startPlayer()  to play a file
- startPlayer() to play from a URI

Actually, file paths and URIs are completely mixed, and I feel that it is a mess.

Larpoux:
- startPlayerFromURI() to play from a URI

(sorry for the typo)

claptimes5:
Yep. No issues with the ":" in the file name. The file plays fine.

claptimes5:
At least, no issues when using the `file://`

hyochan:
@claptimes5 Can we support this optionally? Whether user wants `Uri` or `file path`.

claptimes5:
Sure... though this PR doesn't change your API at all. A user still just provides an absolute path and this PR converts it to a URI so the file can be referenced.

Larpoux:
Actually, we think that modifying the path that the App provide is not a good idea, and could give errors for existing Apps.

claptimes5:
@Larpoux I would like to reopen this issue. I believe this code only provides better support for more file types and should not impact existing apps. Can you let me know if there are issues? I have updated the code to make merging possible for version 3.

Larpoux:
I am sorry, but I do not understand why   opening a file specifying an URI instead of a file path would be better for the OS.
If you really want to pass an URL, then go head : there is no need to patch the flutter_sound code.
FlutterSound do not do anything on what you pass as parameter. So you can pass a URI if really you want that.

One OS  (Android or iOS, I do not remember which) uses a different API verb to access a file via a URL. If we want to support both OS, we must implement a new Flutter Sound verb, so that the App can open its file with a URL. If this is necessary we can consider to do that in a future version.

Larpoux:
If we patch flutter_sound without really understand why we do that, then it can have terrible consequences on other users.
Actually, I do not understand the reason of your patch.

claptimes5:
Thanks for getting back to me. The reason I'm reporting this issue is because the documentation says I can pass a `String path` to a file to `startPlayer`. But when I pass `/storage/emulated/0/Android/data/com.myapp.example/files/entries/entry_2020-04-10T12:01:24.995154.aac` to `startPlayer` it fails to `prepare`. 

That seems like a problem with the `flutter_sound` code, not with my usage.

I'm happy to just pass a URI to `startPlayer` if you don't think this is a good patch.

Thanks for all your work on this!

claptimes5:
Or I think the documentation should change to say something like "If the file fails to play/prepare, try passing a URI parameter."

Issue #234, opened 2020-03-06
[FLAUTO] Infinite loop after pause/resume on flauto


Larpoux:
### Version of flutter_sound
flauto and 2.1.1

### Platforms you faced the error (IOS or Android or both?)
ios

### Expected behavior
Pause and resume should work

### Actual behavior
sometimes the app goes in an infinite loop

### Tested environment (Emulator? Real Device?)
Real ios device and ios emulator

### Steps to reproduce the behavior
pause/resume/pause/resume/pause/resume/pause/resume/pause/resume/pause/resume

This is a strange bug. ios enter a loop where it looks like playing a sound. But in reality nothing is played. The progress bar progress, and when finished, restarts.
Except the progress callbacks, my App does not have control.

Actually I do not have identified if it is : 
- a bug in flauto
- a bug of my implementation of flauto
- a bug inside my App itself

This issue must be understood before releasing flauto.
@salvatore373 : if you have any idea, just tell me 😺

salvatore373:
@Larpoux , I'm sorry but I can't reproduce the issue hitting play and pause repeatedly. Could you please post a video of the actual behaviour and, if anything gets printed, a what the console thinks about it?

Larpoux:
@salvatore373 : I do not think this is **your** job to debug **my** bugs.
But of course, if you can help, you are welcome.
Here is [a video](https://github.com/Canardoux/flutter_sound/blob/flauto/Screen%20Recording%202020-03-08%20at%2018.10.57.mov) Unfortunately, this will probably not be very useful.

salvatore373:
@Larpoux , I think that this problem is not related to my code, because, even creating a single play/pause button to perform a faster repeated switch, nothing occurs in my version of the plug-in. So I think that it is related to your flauto code, and I encourage you to double check it to find the line that causes the issue 😉.

Larpoux:
Hi @salvatore373 , hi @hyochan ,

I reproduce this bug without difficulty on the **flutter_sound example**, and **regular plugin 2.1.1**.
This bug is not related to ```flauto```, but is an ancient bug. A very serious bug.

I do not see any raison to delay anymore the ```flauto``` release because of this bug.
Flauto is not worst that the current flutter_sound actually released.

Now I will 
- check @salvatore373  code
- clean my code, 
- do a beta-test with ```Diglot Korean``` (my app)

and if everything ok, we will be able to release ```flauto``` this week 😃

Larpoux:
[...]

Issue #232, opened 2020-03-04
Operationg multiple flutter_sound instance independently


kiha-la:
### Version of flutter_sound

flutter_sound: ^2.0.4

## flutter doctor

[√] Flutter (Channel stable, v1.12.13+hotfix.8, on Microsoft Windows [Version 10.0.17763.1039], locale ja-JP)

[√] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
[√] Android Studio (version 3.6)
[√] VS Code, 64-bit edition (version 1.42.1)
[√] Connected device (1 available)
    

### Platforms you faced the error (IOS or Android or both?)

Tested on Android

### Expected behavior

1. Call only one `onRecorderDbPeakChanged` callback within the multiple `FlutterSound` instances.
2. Able to stop any recorder one of them.

### Actual behavior

1. `onRecorderDbPeakChanged` for all `FlutterSound` instances.
2. Can only control the recorder that started last.

```
Restarted application in 1,157ms.
D/FlutterSoundPlugin(11332): rawAmplitude: 0.0 Base DB: 0.0
I/flutter (11332): startRecorder: /data/user/0/com.dooboolab.fluttersoundexample/cache/file_dbLevel_disable.aac
D/FlutterSoundPlugin(11332): rawAmplitude: 0.0 Base DB: 0.0
I/flutter (11332):
I/flutter (11332): got dB update of  file_dbLevel_disable -> 0.0
I/flutter (11332): 	2020-03-04T15:24:33.141111
I/flutter (11332): startRecorder: /data/user/0/com.dooboolab.fluttersoundexample/cache/file_dbLevel_enable.aac
D/FlutterSoundPlugin(11332): rawAmplitude: 1410.0 Base DB: 42.67625931342095
I/flutter (11332):
I/flutter (11332): got dB update of  file_dbLevel_disable -> 42.67625931342095
I/flutter (11332): 	2020-03-04T15:24:34.111216
I/flutter (11332):
I/flutter (11332): got dB update of  file_dbLevel_enable -> 42.67625931342095
I/flutter (11332): 	2020-03-04T15:24:34.111296
D/FlutterSoundPlugin(11332): rawAmplitude: 1175.0 Base DB: 41.09263439246845
I/flutter (11332):
I/flutter (11332): got dB update of  file_dbLevel_disable -> 41.09263439246845
I/flutter (11332): 	2020-03-04T15:24:35.112086
I/flutter (11332):
[...]

PR #243, opened 2020-03-15
[FLAUTO]

Larpoux:
OK, guys and girls.
I think I am ready for an integration inside the master branch.
The only, **BIG** issue not fixed is #234 . This issue is very, very serious. I had to disable the `Resume` feature on `Diglot Korean` (my own App). This is really very bad 😞 .
But I think this bug is not related to Flauto, and is already present in the actual Master branch.

@salvatore373 : Can you look to this Pull Request and tell us if you have remarks ?
@hyochan : I hope that you will like this PR and integrate it soon in the Master Branch.


Larpoux:
@hyochan : 
Now, flutter_sound has two different modules:
- The legacy `flutter_sound` with some improvments like some tools to control the Audio-focus 
- `Flauto`, wich adds @salvatore373 improvements .

Those two modules are completely separated. This has two benefits:
- Separated functionalities means easier maintenance
- The legacy `flutter_sound` module is totally backward compatible

My project would be to split again the legacy `flutter_sound` module into two different modules : 
- FlutterSoundRecorder
- FlutterSoundPlayer
Those two functionalities are not related together, and there is no reason to have just one module.
Having two smaller modules will let the maintenance easier.

For having backward compatibility, the `FlutterSoundPlayer` module will be called simply `FlutterSound` and will inherit from `flutterSoundRecorder`.
But new developer will be encouraged to use the `FlutterSoundRecorder` module for recording.

We will have three modules : 
- `Flauto` (which inherit from `FlutterSoundPlayer)
- `FlutterSoundPlayer` (which inherit from `FlutterSoundRecorder` for backward compatibility)
- `FlutterSoundRecorder`

Then, I think it will be easier to address #232 (which has been closed by its creator, but for me this issue is still valid and interesting).

hyochan:
@Larpoux @salvatore373 Thanks for your time for bringing up fabulous features. I really appreciate all your work given to `flutter_sound` for a long time and I'll do my best to check this `PR` so that your times are not wasted. I will try to finish this review in March!

Thanks again to all you wonderful people 🥇

Larpoux:
Hyo, I suggest to maintain two branches : 
- `Stable` (the actual Master branch)
- `Master` (the Flauto branch)

and merge the master branch to the stable branch if/when the master branch is accepted by the community as stable.

bsutton:
Can I suggest you consider renaming the Flauto module to something like FlutterSoundPlayerV2.

The name flauto doesn't convey any useful information. 

If we use my suggested naming convention then we end up with
FlutterSoundPlayer
FlutterSoundPlayerV2
FlutterSoundRecorder

Selecting the appropriate module is now easy.

Larpoux:
> Hyo, I suggest to maintain two branches :
> 
> * `Stable` (the actual Master branch)
> * `Master` (the Flauto branch)
> 

@hyochan : what do you think of this ? I think it will not give us much more maintenance work.
And it will be easier for us to split the Problem Reports between 'bug in Master Branch' and 'bug in stable branch'.

I am ready to do some maintenance and development on the `Master Branch`.
It will be more comfortable, for me to know that I will no break the `Stable Branch`.
If we do that, maybe good if I accept your invitation to be a flutter_sound maintainer. 😸

Larpoux:
@hyochan is actually very, very busy for flutter_sound maintenance.
In the interim I suggest that we use the `flauto` fork.

The sources [are here](https://github.com/Canardoux/flauto).
- branch `stable` is the actual flutter_sound (V 2.1.1). This is the **gamma** branch.
- branch `beta` is the new plugin with advance features (V 3.0.1). This is the **beta** branch.
- branch `master` is just for the project development and not to be used. This is the **alpha** branch.

The flauto plugin is published. In your pub spec.yaml : 
```
dependencies:
  flutter:
    sdk: flutter
  flauto: ^3.0.1+1
```

hyochan:
@Larpoux Sounds nice that some of them can try newer version beforehand. I like the idea of separating the branches too. Shall we test the `beta` release for this plugin and further review this ongoing?
If you separate the `pub` plugin to `flauto`, users may get confused about what plugins to use since they are not always focusing on our work. What about just releasing this to the `beta`? How do you think?

hyochan:
Merged 🎉

hyochan:
Released 3.0.0 🎉

Issue #251, opened 2020-03-24
Feature request, decibel meter sample time


lukepighetti:
Hello! I am trying to capture loud noises (clipping) with a 1/100 sec precision and a 1/10 sec refractory period. I've been able to capture these noises at a rate close to 1/100 sec and a refractory period of 1/2 sec with this project. Do you have any guidance for how we can improve the performance for my use case?

TLDR: I would like to be able to set the decibel meter sample time to 0.05 sec. ie. `FlutterSound.setDbPeakLevelUpdate(intervalTime: 0.01, sampleTime: 0.05)`

lukepighetti:
Handled by https://github.com/dooboolab/flutter_sound/pull/252

Larpoux:
Is your request for iOS or Android ?
Flutter Sound 3.1.6 changed a little bit its code for iOS (#252).

Could you try, and tell us if your request is completely addressed ?

PR #252, opened 2020-03-25
setAudioMeterToAmbient

lukepighetti:
This allows the package consumer to immediately set the audio meter to ambient, iOS only implementation so far.

Addresses https://github.com/dooboolab/flutter_sound/issues/251

Apologies for the large diff, I didn't know this package wasn't using `dartfmt`

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -8,12 +8,15 @@ PODS:
   - mobile-ffmpeg-audio (4.3.1)
   - path_provider (0.0.1):
     - Flutter
+  - path_provider_macos (0.0.1):
+    - Flutter
 
 DEPENDENCIES:
   - Flutter (from `Flutter`)
   - flutter_ffmpeg/audio (from `.symlinks/plugins/flutter_ffmpeg/ios`)
   - flutter_sound (from `.symlinks/plugins/flutter_sound/ios`)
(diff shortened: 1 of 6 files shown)


Larpoux:
This pull request seems interesting.
But I wonder what to do for Android ?

hyochan:
@lukepighetti Same question as @Larpoux

lukepighetti:
If I'm being honest I have no idea what to do on android or if it's even necessary. I know that's not helpful but that's what I've got at this moment.

lukepighetti:
Looks like I won't be getting around to nailing this down on Android. I won't be upset if you close this out. Thank you for your time.

Larpoux:
No, do not close this issue until we look to that seriously.
But time is really a problem. I have too many things in the pipeline...

Larpoux:
Hi @lukepighetti ,

I am looking to your Pull Request. 
I am wondering if instead for adding a new API verb, it would be good to always do a ```isMeteringEnabled: true``` when the App call the API verb ```setDbLevelEnabled(true)```

I guess that if the App wants to control the amplitude meter, this one must be accurate, even if it means some CPU eating.

Larpoux:
We added the iOS call to ```setMeteringEnabled:YES``` in ```setDbLevelEnabled(true)```.

PR #268, opened 2020-04-04
Fix error on documentation for Android FFmpeg setup

oliversd:
The README file says that you should enter
ext.flutterFFmpegPackage = 'audio-lts' on the pubspec.yaml,
when in reality, it should be into the android/gradle.build file.


Larpoux:
Oh !  Yes, you are right.
We are going to fix that.
We will merge your Pull Request in next release.

oliversd:
Great! Thank you

PR #274, opened 2020-04-07
flixed conflic with permission_handler 5.x.x

darwin-morocho:
I have updated the startRecorder method to be compatible with permission_handler: 5.x.x. And removed unnecessary imports.


Larpoux:

Do you think that flutter_sound could be compatible both with version 5.x.x and 6.x.x ?

Larpoux:
I will check your Pull Request tomorrow. Have a good evening/morning/night/day .

darwin-morocho:
> Thank you for your contribution.
> 
> Do you think that flutter_sound could be compatible both with version 5.x.x and 6.x.x ?

that will be compatible just with versions 5.x.x.  maybe for 6.x.x the permissions_handler plugin could be have significant changes.

PR #279, opened 2020-04-09
added pedantic lints, major refactoring of example with bug fixes.

bsutton:
I've added pedantic lints to the code based and fixed all the resulting lint errors.

I've also done a major refactor of the example to help improves its long term maintainability.

Included are a number of fixes for the examples UI.


Larpoux:
Thank your @bsutton : at first seeing it seems to be very good.
I will look to your request in a few days.
thanks again.

Larpoux:
The instance variable ```trackPlayer``` does not seem to be initialized.

```

/// This class is deprecated. It is just to keep backward compatibility.
/// New users must use the class TrackPlayer
@deprecated
class Flauto extends FlutterSound {
  TrackPlayer trackPlayer;
  Flauto() {
    initializeMediaPlayer();
  }

  void initializeMediaPlayer() async {
    if (soundPlayer == null) soundPlayer = TrackPlayer();
    if (soundRecorder == null) soundRecorder = FlutterSoundRecorder();
    await soundPlayer.initialize();
    await soundRecorder.initialize();
  }

  Future<String> startPlayerFromTrack(
    Track track, {
    t_CODEC codec,
    TWhenFinished whenFinished,
    TwhenPaused whenPaused,
    TonSkip onSkipForward = null,
    TonSkip onSkipBackward = null,
  }) async {
    return await trackPlayer.startPlayerFromTrack(
      track,
      whenFinished: whenFinished,
      onSkipBackward: onSkipBackward,
      onSkipForward: onSkipForward,
    );
  }
}
```

Larpoux:
I have several problems with your new example : 

- When a playback is finished, the icons "Pause/resume" And "Stop" are still enabled, and the icon "Play" is still disabled.

- Play from Asset does not work

- The switch "Allow Tracks" and "Duck Others" are always disabled.

- After playing a Remote Example, I was not able to recover an enabled icon "Record" after returning back to "Media:File" and "Codec:AAC"

Larpoux:
In track_player.dart : ```problems in ```channelMethodCallHandler``` : 
```
.audioPlayerFinished(call.arguments as Map<String, dynamic>);
```
Does not work. I had to remove the type casting.

bsutton:
> In `track_player.dart` : problems in `channelMethodCallHandler` :
> 
> ```
> .audioPlayerFinished(call.arguments as Map<String, dynamic>);
> ```
> 
> Does not work. I had to remove the type casting.

I've resolved this in the latest pull request.

bsutton:
> I have several problems with your new example :
> 
> * When a playback is finished, the icons "Pause/resume" And "Stop" are still enabled, and the icon "Play" is still disabled.
Resolved
> * Play from Asset does not work
Resolved
> * The switch "Allow Tracks" and "Duck Others" are always disabled.
The Switch 'look' grayed out but you can 
> * After playing a Remote Example, I was not able to recover an enabled icon "Record" after returning back to "Media:File" and "Codec:AAC"

This is part of a broader problem with the entire example UI. For instance you shouldn't be able to change Media or Codec whilst recording.  To resolve these issues we really need to have a global controller which tracks the state of the recorder and player so that we can disable components and re-enable them after action completes.
I don't have time to address this at the moment but I don't think the changes I've made have made the situation any worse (these class of issues were pre-existing).

Larpoux:
I had an exceptions during tests on the Example App
Screenshot 2020-04-14 at 11 09 29

Larpoux:
'Allow tracks' and 'Duck Others' are disabled on the Example App.

Larpoux:
When there is nothing to play, the 'Playback' button is not disabled (but there is a Toast Message, which is good).
Is it what you want ?

Larpoux:
I created a new branch '4.0.0-beta.1', which include your Pull Request.
I would like to change the API a little bit : 
- Remove the deprecated classes
- Rename Public enums like 't_CODEC' to be conform with Upper Case Camel Recommendations.
- Rename 'whenFinished:' by 'onFinished:' to be homogeneous with 'onSkipForward:'

This new version 4.0.0 will support PCM recording both on Android and iOS.
Maybe it will get rid of the flutter_ffmpeg depency

Please, do your future pull requests on this branche.

Larpoux:
Merged into '4.0.0-beta.1'

bsutton:
Ok thanks

bsutton:
Correct. The toast let's the user know how to fix the problem rather than
leaving them guessing.

Larpoux:
That sounds good to me. In general I do not like when Widget are disabled when no clue for the user to understand why

---------
/Larpoux

Larpoux:
And probably good to allow the user to play-back and record at the same time (Flutter Sound is now supposed to handle that correctly).
But I got several bugs when doing a specific sequence of actions.
I have not yet determined which sequence. (I am currently working on V 3.x.x maintenance)

---------
/Larpoux

Larpoux:
On Version beta.1 : 
- the two switchs are always disabled.
- After playing OPUS the 'Playback' button stays disabled. (Perhaps a regression during my last merge ...)

bsutton:
To allow a user to record and playback simultaneously via the example would
require a rework of the UI.

I don't think the example needs to be overly complicated.

At some point we are probably better off having a couple of very simple
examples.
One that records and one that plays back.

The idea is to give enough code so a user can just pick up an example and
cut and paste the pieces.

I've got a couple of controls that I might be able to donate. A playbar and
a recording bar that are meant to be dropped in. I don't have the time at
the moment but something I might be able to do over the coming months.

bsutton:
> On Version beta.1 :
> 
> * the two switchs are always disabled.
> * After playing OPUS the 'Playback' button stays disabled. (Perhaps a regression during my last merge ...)

I can't reproduce the problem with the switches being disabled.

As to OPUS I can't test this as I only have access to an android phone at the moment.
I'm guess some exception gets thrown can you do a test run with the debugger set to catch all exceptions and see what bubbles out.

By the way: what does Duck Other do?

Larpoux:
I agree that the example(s) mus be very simple.
We are not developing a real App, but something to show how Flutter Sound API can be used.

Actually our Example is more a ```Demo``` than an ```Example```.
But maybe there is place for Example(s) and a demo.

Larpoux:
When we play something during Spotify already playing some music, we can specify what to do : 
- Stop Spotify
- Play both our sound and Spotify
- Or lower Spotify Sound during our playback. This is Duck Others.

Larpoux:
> I can't reproduce the problem with the switches being disabled.
> 
> As to OPUS I can't test this as I only have access to an android phone at the moment.
> I'm guess some exception gets thrown can you do a test run with the debugger set to catch all exceptions and see what bubbles out.

Ok, I will take care of that.
By the way, thank you for your contribution 👍

bsutton:
Agreed. The existing 'example' is really a demo.  We need separate simple
examples.

bsutton:
Then I'm still confused.

We have a toggle switch for duck others but you seem to be saying that
there are three options?

PR #293, opened 2020-04-16
Minor cleanup of lints and made some methods private.

bsutton:


bsutton:
> Thank you very much @bsutton for your help to maintain flutter_sound.
> This is great that you started to add comments missing and cleaning the code.
> I think That we should maintain 3 branches :
> 
> * The release Version (3.x.x). This branch will be published. This is the Master Branch
> * A beta version (4.x.x-beta). This branch will be published and will be tried by some users.
> * A dev version which will not published but maintained on github

We can actually push a dev version as well if we think it would be useful.

I'm currently working on a major restructure along the lines that I've suggested in the 4.0 issue discussions.
I should be able to push something in the next 24 hours that we can have a look at.
I've also been restructuring the code to bring it in line with googles standards (e.g. most code in the src directory).

Larpoux:
Yes, I think this is not good to develop on a branch which is released or in beta-state.

- Branch Beta is supposed to be merged from master, when bugs are fixed, or when small improvements have bee made.
- Branch Beta is supposed to be merged from dev, when we think that dev is stabilized and is able to be tested by some developers.

This means some work for me to maintain 3 branches, but the result will be more secured.
You will probably do your Push Request on dev, I gess.

With the `example`, I have some concerns : actually I am not convinced by the Pull Request that I merged to branch `dev` : 
- My feeling is that there are too many sources files for just a demo.
- I am really not convinced that it is better to enable buttons which may not be used.
- I have several crashes when we start a record, and start at the same time a playback, and then pause something (I do not remember if it is the Play pause or the Record Pause)

bsutton:
> * My feeling is that there are too many sources files for just a demo.
The no. of source files really has really just highlighted the existing complexity of the example.
With few exceptions I've really just split out classes that already existed to make them more obvious and to reduce the complexity in each file.
Its good code practice to keep files less than 500 lines and 300 is better.

I would like to see further work on the example and the development of some discrete and reusable widgets which would become part of the overall package. I've mentioned in another post that I have some widgets that whilst needing some work might fit this requirement.

> * I am really not convinced that it is better to enable buttons which may not be used.
This is really about user discoverability. When I first tried out the Flutter Sound example I couldn't work out how to use it. Enabling these buttons make it much easier to work out how to use the system because it now tells you what the problem is.
I'm not religious on this point but from my experience in usablity labs the current design will get a better reception from users.

> * I have several crashes when we start a record, and start at the same time a playback, and then pause something (I do not remember if it is the Play pause or the Record Pause)

Not surprised given the amount of rework. Send me the stack traces and any details you have on how to reproduce them and I will have a look at them.

Larpoux:
> I would like to see further work on the example and the development of some discrete and reusable widgets which would become part of the overall package. I've mentioned in another post that I have some widgets that whilst needing some work might fit this requirement.

That will be great.
Actually, on dev, I kept the old example and your new proposition.
But I have some problems with Android Studio when there are several 'examples app` in the plugin directory.
The `auto completion` feature, for example does not work any more. Neither the popup helps/hints. Probably something stupid in my parameters. But I had exactly the same problems on other projects I work on.

PR #301, opened 2020-04-17
api rework

bsutton:
So this is a very large pull request and I've only done superficial testing so far.

The pull requests reflects my suggestion to simplify the external api to flutter sound.

The core piece is that I've merged TrackPlayer and FlutterSoundPlayer into a single class SoundPlayer.
I've also renamed FlutterSoundRecorder to SoundRecorder.

A user can now simply construct a SoundPlayer and call the appropriate method:

SoundPlayer.startPlayer();
SoundPlayer.startPlayerFromTrack();

As per googles recommendation I've moved all files into a src directory and have a single package library at the top (flutter_sound.dart) which selectively exposes the parts of the api which should be public.

To reduce the public api I've used Proxies between the plugins and the player and recorder. Without the Proxy we were exposing low level methods such as invokeMethod.

I've rationalised the management of dataBuffers as the orginal code had both the java/swift code writing buffers to files as well as the dart code. I've not touched the java/swift code but I'm now marshalling all buffers into temp files before calling into the plugin. I believe we are better off doing as much code in dart as its easier to maintain and reduces replication of effort (e.g. any code done in the plugin has to be done twice).

I've removed large amounts of duplicated code and split the classes out into separate dart files which should make the code easier to maintain and understand.

This change does break a chunk of the API but given we are moving to 4.0 with the intent of breaking the API it seems best to break it once in a major way and set ourselves up to minimise future breakage.

I will do additional testing over the next week and submit patches as I fix bugs.

Have a look at the structure and api I've setup and lets have a discussion.
I know these changes are significant ( I usually try to keep my PR's to more manageable bits) but we really needed a re-org hence this fairly significant PR.

Brett


Larpoux:
Wow ! You really did much work.
I am going to look deeply on this PR.

For now : 
> The core piece is that I've merged TrackPlayer and FlutterSoundPlayer into a single class SoundPlayer.

I will look to your PR, but actually I am not really sure that it is a good idea

> I've also renamed FlutterSoundRecorder to SoundRecorder.

👍 


> As per googles recommendation I've moved all files into a src directory and have a single package library at the top (flutter_sound.dart) which selectively exposes the parts of the api which should be public.

👍 

> To reduce the public api I've used Proxies between the plugins and the player and recorder. Without the Proxy we were exposing low level methods such as invokeMethod.

👍 

> I've rationalised the management of dataBuffers as the orginal code had both the java/swift code writing buffers to files as well as the dart code. I've not touched the java/swift code but I'm now marshalling all buffers into temp files before calling into the plugin. I believe we are better off doing as much code in dart as its easier to maintain and reduces replication of effort (e.g. any code done in the plugin has to be done twice).

👍 

> I've removed large amounts of duplicated code and split the classes out into separate dart files which should make the code easier to maintain and understand.

👍  : This was something that I wanted to do myself. Thanks to did it

> This change does break a chunk of the API but given we are moving to 4.0 with the intent of breaking the API it seems best to break it once in a major way and set ourselves up to minimise future breakage.

Yes, I think it is the good moment to clean all the API on 4.0.x


> I know these changes are significant ( I usually try to keep my PR's to more manageable bits) but we really needed a re-org hence this fairly significant PR.

I really appreciate your work and what you are doing.
I realize that adding many comments missing, getting rid of all the warnings, and change the code so that it will follow Google recommendations is really a thankless job.
So thank you to spend some of your time on that.

Larpoux:
Hi, Brett.
I merged the 'dev' branch into the 'beta' branch, just before including your Pull Request because it was the last moment where the API was not changed.
Probably the actual 'beta' branch will be soon '3.2.0' .

I have not yet looked to your PR. I have too much job to do on the 'beta' branch and the 'master' branch. 
I just compiled the example (demo) App. Several problems, specially on Android. 
I am not sure to like very much the new Demo App.

bsutton:
I've not spent much time testing as yet so the demo app is fairly broken.
I will be working through fixes over the next week.

What tasks are you working on?

Larpoux:
Maintenance, Maintenance, ... and Maintenance.
Last few days, we had many new issues.
- Some developers just need help
- Some developers want improvement/enhancement
- And some users get crashes in TrackPlayer.

I worry a little bit with TrackPlayer. Salvatore did some code too complicated for my taste, and hard to debug because most tasks are done asynchronously and called from system callback.
I would not want that Track Player be a bug nest.

bsutton:
Yes I also thought the TrackPlayer was overly complicated at the dart
level.
That was why a big chunk of my code changes was simplifying the
implementation.
There were large amounts of duplicated code and dependencies.
I think the implementation I've pushed in the PR is easier to understand
and should make it easier to maintian.

PR #305, opened 2020-04-18
Bug fixes

bsutton:
Looks like all the playback options are now working.
There is an issue with the recording icon staying grayed out after doing a playback which I'm looking into.

I've also raised an issue regarding slot management which is a problem with hot reloads.

This is now looking fairly good.
Perhaps need some work to make the Playbar UI prettier.
We could also add a switch in the example to control whether the playbar displays the track title or note.


bsutton:
Merged after pull and removed lots of lints.

bsutton:
So this is now feeling fairly robust.

Work to do:
Test Hush - I have audio issues on my test env at the moment so haven't been able to test.
refactor the SoundRecorder API so it uses the same style interface as SoundPlayer.

Larpoux:
Could you explain me why you have two Push Requests : 
#302 and #305 ?

bsutton:
this pull request was against the wrong branch.

bsutton:
Looks like I made a pull request against the wrong branch. I've deleted the
pull request against master and left the one against dev.

PR #311, opened 2020-04-19
Improved documentation for Playbar.

bsutton:
Also an npe fix and better formatting.


bsutton:
I've done some initial work on the READ.ME as I think it will help clarify the simplicity of the new api.

bsutton:
This update includes an prototype for an Album api that would allow you to play a collection of tracks on the OS UI.


I've updated the read.me to explain how the api works.
We now have:
SoundPlayer - used to play audio with no ui.
Track - play audio via the UI.
This differs from the original TrackPlayer/Track in that it simply uses a SoundPlayer rather than inheriting from it which provides a cleaner code base. If you look at the Track implementation its quite tiny and almost complete.
Album takes a set of Tracks either static or virtual.
The virtual version allows you to dynamically create an album.
The album delegates playing the audio to SoundPlayer which again makes the Album implementation clean and simple as it only has to deal with changing tracks.

To make this work neatly we need to merge the plugin TrackPlayer and FlutterSoundPlayer so we can connect the SoundPlayer to a single plug and the dynamically choose which player to use.
Look at the code I don't think this will be too hard to do.

PR #312, opened 2020-04-20
Album api prototype

bsutton:
This update includes an prototype for an Album api that would allow you to play a collection of tracks on the OS UI.

I've updated the read.me to explain how the api works.
We now have:
SoundPlayer - used to play audio with no ui.
Track - play audio via the UI.
This differs from the original TrackPlayer/Track in that it simply uses a SoundPlayer rather than inheriting from it which provides a cleaner code base. If you look at the Track implementation its quite tiny and almost complete.
Album takes a set of Tracks either static or virtual.
The virtual version allows you to dynamically create an album.
The album delegates playing the audio to SoundPlayer which again makes the Album implementation clean and simple as it only has to deal with changing tracks.

To make this work neatly we need to merge the plugin TrackPlayer and FlutterSoundPlayer so we can connect the SoundPlayer to a single plug and the dynamically choose which player to use.
Looking at the code I don't think this will be too hard to do.


bsutton:
Worth noting that the concept of an Album will help those that want us to cache audio.
Wrapping a virtual Album in a cache would allow the cache concept to implemented fairly easily and provides a coherent scope for the life time of the cache.

Larpoux:
```
/// TODO
  /// Is this in the correct spot if it is only called once?
  /// Should we have a configuration object that sets
  /// up global options?
  Future<bool> iosSetCategory(
 ```

```
/// TODO
  /// Is this in the correct spot if it is only called once?
  /// Should we have a configuration object that sets
  /// up global options?
  Future<bool> androidAudioFocusRequest(int focusGain) async {
 ```

I think that we must add some parameters during ```initialize()``` for specifying things like "Duck Others" or if the App do not want to acquire the Audio Focus automatically, and wants to manage the Focus itself.
Those parameters would be High level and OS independent.

Also, for backward compatibility with the legacy code, actually by default I request the Audio Focus when doing a startPlayer(), and abandon it when stopPlayer().
1. The code is really dirty (I still have not looked to your new code ...)
2. By default, I think better to acquire the audio focus when opening the session, and keep it until release().

I think that the two verbs above should be kept just in case the App wants to deal directly with the OS, but will not be used in a regular App.

About the Audio Focus : 
- iOS uses the API verb : `setActiv(true/false)`
- Android uses the API verb : `RequestAudioFocus` and `AbandonAudioFocus`

I choose to use the iOS terminology , but I think now that it was a bad choice.

Larpoux:
About Codec : 

Actually, the Codec enum mix two concepts : 
- The codec itself
- The file format

Theoretically, when recording a sound, the user should be able to specify both the codec and the file format. But audio world is really a mess. 
There are only a few combination allowed. For example it is not possible to store an Opus Record into a WAV file format.
More, most of the users does not understand those two concepts. They probably want to record a MP3, or a VORBIS without dealing with those two concepts.
Also, it would be very difficult to implement the verb `isCodecEncoderSupported()`, because we would have to take care of which combinations are allowed.

So, I think that mixing the two concepts (as we are currently doing) is correct.

bsutton:
> About the Audio Focus :
> 
> iOS uses the API verb : setActiv(true/false)
> Android uses the API verb : RequestAudioFocus and AbandonAudioFocus

I too prefer the term 'AudioFocus' which is what I think you are saying.

Do we have requestAudioFocus and abandonAudioFocus 
or 
setAudioFocus(true/false)

Larpoux:
Not very important. Probably "RequestAudioFocus()" and "AbandonAudioFocus()" is clearer.

bsutton:
> Also, for backward compatibility with the legacy code, actually by default I request the Audio Focus when doing a startPlayer(), and abandon it when stopPlayer().
> 
> The code is really dirty (I still have not looked to your new code ...)
> By default, I think better to acquire the audio focus when opening the session, and keep it until release().

We should really let the user control the focus.
I can see the logic on grabbing the focus when you call startPlayer but I can see circumstances where the users really doesn't want to grab the focus.

I will leave it to grab the focus for the moment but I suspect that its something we need to come back to.

bsutton:
FYI: I'm doing a major refactor of the code to support Albums so for the moment I would suggest that you hang of doing changes as the merge might be difficult.

bsutton:
Did you have a chance to look at the read.me and the doco on Album and Track?

Larpoux:
> Did you have a chance to look at the read.me and the doco on Album and Track?

Not yet.



> FYI: I'm doing a major refactor of the code to support Albums so for the moment I would suggest that you hang of doing changes as the merge might be difficult.

Yes, I agree. Better that you are alone to modify the code , actually.
I am going to go back to my other projects which are now **very very** late.

bsutton:
will do.

bsutton:
How does the method androidAudioFocusRequest

relate to setActive?

Larpoux:
On Android, the App specifies what to do with others when it Requests the Audio Focus.
On iOS, the App specifies what to do with others for the Audio Session duration, and just specify `setActive(true)` when it wants the Audio Focus.

To be iOS independent, I choose the iOS way for handling the Audio Focus.
And this simplify what we must do when the App startPlayer() but did not specify anything for what it wants to do with the Focus

bsutton:
What is the difference between duck others and request focus.

I thought that both would do the same thing?

bsutton:
Does Focus attach the audio to the is is ui?

Larpoux:
> What is the difference between duck others and request focus.

When we open a session, we specify the parameters of the session.
For example what to do when we aquire the an Audio Focus, if the session is for playing ambient music, if it is to play speech, ...

`SetActive()` is just request to get the Focus

bsutton:
I'm not quite understanding your answer.

I assume that if you set duck others then any audio playing from other
sources will have its volume reduce.

What I'm confused with is setActive? Does setActive change the volume or
does it open the OS media player or something else?

bsutton:
So I've been looking at the plugin and there are two very similar methods
in the plugin:

androidAudioFocusRequest
which calls:
audioFocusRequest = new AudioFocusRequest.Builder ( focusGain )
and
requestFocus
which calls
audioFocusRequest = new AudioFocusRequest.Builder (
AudioManager.AUDIOFOCUS_GAIN )
and then check if focus was granted.

Why are there two almost identical methods?

Larpoux:
I tried to emulate the iOS processing in Android, so that `SetActive()` will be OS independent.
`androidAudioFocusRequest` is not OS independent.

bsutton:
Reading the flutter sound doco it states that:
androidAudioFocusRequest should be called once on app startup.

Given it appears identical to the 'requestFocus/setActive' method (with the
exception that it doesn't check if the grant was successful)  I don't think
this method is necessary.
The google docs say you must request focus on the callback to play which I
can see the plugin does.
So I can't see what part this method is meant to play.

bsutton:
OK I see the difference now. the androidAudioFocusRequest method allows you
to pass a 'gain'.
Looks like the doco is wrong in saying it should be applied when the app
starts given focus is lost when playback stops.
The code also calls it each time startPlayer is called.

bsutton:
So I think I'm now down with the major code refactoring.
The only significant piece is the recorder which I will do tomorrow.

There has been zero testing on this code base which I will start on once the recorder is complete.

I've updated the read.me with the latest revisions of the api as well as a draft of the recording api.

I think the api now provides a balance between competing interest :)

The new AudioSession allows you to maintain a session across multiple files whilst the SoundPlayer provides a simple interface for playing a single audio file.

The TrackPlayer is gone as its no longer needed.

We still have a Track but this can now be used with the AudioSession, Album, SoundPlayer and the SoundRecorderUI widget (previously Playbar).

You can now decide to show the OSs UI for all of the playback methods by passing the correct type of AudioSession. AudioSession.withUI() or AudioSession.noUI().

We now have an orthogonal api (all the bits work with the other bits) which provides a consistent experience for users. [e.g. I can use a Track with the SoundPlayer and with the Album].

The api also gives users a nice progression from 'I just want to play a file' via SoundPlayer to 'I want to roll my own media player via AudioSession'.

Yes there are a lot of changes here but I think we have an api that sets flutter sound up for the future.
Its flexible and easy to use and we now have a reasonable amount of documentation (with more coming).
From my experience with other open source projects if you can get the documentation and ease of use right, you are 80% of the way to a successful project.

bsutton:
All of the noted api changes.
Updated the read.me to reflect the new apis

Still have recording to do. Off to do that now.

Issue #307, opened 2020-04-19
Possible race condition: NullPointerException: Attempt to invoke virtual MediaControllerCompat


bsutton:
I'm unable to reproduce the problem but lodging this incase it reoccurs and we have some additional information.

I produced this originally by:
Start example
Played MP3 Asset twice
Enabled Show on OS UI
Played MP3 asset twice

The following error was thrown:
```
W/MediaPlayer(26404): Use of stream types is deprecated for operations other than volume control
W/MediaPlayer(26404): See the documentation of setAudioStreamType() for what to use instead with android.media.AudioAttributes to qualify your playback use case
E/MethodChannel#com.dooboolab.flutter_sound_track_player(26404): Failed to handle method call
E/MethodChannel#com.dooboolab.flutter_sound_track_player(26404): java.lang.NullPointerException: Attempt to invoke virtual method 'android.support.v4.media.session.MediaControllerCompat$TransportControls android.support.v4.media.session.MediaControllerCompat.getTransportControls()' on a null object reference
E/MethodChannel#com.dooboolab.flutter_sound_track_player(26404): 	at com.dooboolab.fluttersound.TrackPlayer.startPlayerFromTrack(TrackPlayer.java:331)
E/MethodChannel#com.dooboolab.flutter_sound_track_player(26404): 	at com.dooboolab.fluttersound.TrackPlayerPlugin.onMethodCall(TrackPlayer.java:152)
E/MethodChannel#com.dooboolab.flutter_sound_track_player(26404): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:231)
E/MethodChannel#com.dooboolab.flutter_sound_track_player(26404): 	at io.flutter.embedding.engine.dart.DartMessenger.handleMessageFromDart(DartMessenger.java:93)
E/MethodChannel#com.dooboolab.flutter_sound_track_player(26404): 	at io.flutter.embedding.engine.FlutterJNI.handlePlatformMessage(FlutterJNI.java:642)
E/MethodChannel#com.dooboolab.flutter_sound_track_player(26404): 	at android.os.MessageQueue.nativePollOnce(Native Method)
E/MethodChannel#com.dooboolab.flutter_sound_track_player(26404): 	at android.os.MessageQueue.next(MessageQueue.java:336)
E/MethodChannel#com.dooboolab.flutter_sound_track_player(26404): 	at android.os.Looper.loop(Looper.java:174)
[...]

PR #316, opened 2020-04-22
Mostly working code.

bsutton:
So I think I'm finished.

I'm still going to drop a new recording widget into the project that will become part of the public ui to make it easier for users to get up and running.

What we need now is some people trying this out.
Can you guys give it a try and then we can look perhaps at pushing it to pub.dev as a dev build.

I'm fairly happy with where it is now and I believe I've fixed a number of bugs that existed in 3.x

My main concern now is that we aren't a good citizen in that I don't think we play well with other players.

We also probably need to look at the events coming up from OS code such as start/stop/pause/resume events so we can process these state changes in a consistent manner.
At the moment its different depending on whether we use the track player or the sound player plugin.

I also need someone to tweak the iOS plugin to emit onConnect events as the TrackPlayer does for android.

Anyway have a look and give me some feed back.


bsutton:
I should note that I've updated the read.me with details on the api changes.

I've included a migration guide from 4 to 5.

If you could have a read and give any feed back it would be appreciated.

bsutton:
I believe that I've now fixed the onConnection issue.

This was more complicated than thought. The original technique using a CountDownLock failed
as it looks like the wait was blocking the plugins message pump so the onConnect would never complete.
To resolve this I had to add an onConnect callback from the plugin into the dart code.
Prior to this change I could reproduce the issue about every fourth attempt. I'm now unable to reproduce it.

However I need someone to create the onConnect events for the apple code.
I still need to add the onConnect to the non-tracked player for android but for the moment I have a hack in place to simulate the event.
I will fix that hack tomorrow.

I think I also need to look at the logic that start an audio file playing as it appears to be locking up the ui. I suspect that I need to employ the same technique as above (i.e. send a callback).
I now done a fair bit of testing of the example under android and its looking fairly stable. Still need to try a few more combinations and push it a bit harder.

Larpoux:
Is this #317 and #307 ?
If yes, is it possible to apply a patch not too complicated on 3.x ?

bsutton:
So I think I'm finished.

I'm still going to drop a new recording widget into the project that will become part of the public ui to make it easier for users to get up and running.

What we need now is some people trying this out.
Can you guys give it a try and then we can look perhaps at pushing it to pub.dev as a dev build.

I'm fairly happy with where it is now and I believe I've fixed a number of bugs that existed in 3.x

My main concern now is that we aren't a good citizen in that I don't think we play well with other players.

We also probably need to look at the events coming up from OS code such as start/stop/pause/resume events so we can process these state changes in a consistent manner. 
At the moment its different depending on whether we use the track player or the sound player plugin.

I also need someone to tweak the iOS plugin to emit onConnect events as the TrackPlayer does for android.

Anyway have a look and give me some feed back.

Larpoux:
Thank you Brett.

I will do those steps, in this order : 
- Review the dart code, and specially the API
- Review the example (demo) code
- Try to find a good compromise together for points that I do not agree with you
- Do extensive tests, mostly using the Demo App (or patched Demo App)
- Review the documentation (README)
- Review the iOS code
- Review the Android code

Actually I do not have much time. I have to maintain the released version of flutter_sound, and progress on Dyglot Korean (my own App).
For past weeks I spent to much time on Flutter Sound and I am very, very late on Dyglot Korean.

bsutton:
Sounds good.

Fyi: my work is happy to provide on going support for the dart code provide
you can do the maintenance on the iOS Android code.

Hopefully with the combined resources we can make flutter sound the best
sound package for flutter.

Larpoux:
Yes. My thinking is that flutter_sound has the capability to be **THE** flutter plugin for sound playing and recording.
The foundations are very good, thanks to @hyochan who did a great job.
He built a very simple package. I hate complexity, because of the maintenance.
We definitely must keep this product simple.

Larpoux:
Was closed by mistake

Larpoux:
Was closed by mistake

bsutton:
A large part of the work I've done is around making the API simple for
users of the API.

Part of this has been making the API consistent which I think I've achieved.

The next issue was to make certain the code works correctly.

One issues with the current code base was that it was overly simplistic to
the point that it wouldn't do the required job.

In testing of the original code I found an number of non trivial bugs that
were a result of the architecture not being adequate.

The old API also locks down the flutter ui causing frame drops due to
incorrect usage of awaits.
There are still some pieces that use awaits incorrectly but I've left them
for them moment as they don't normally cause frame drops.

I guess my point is that the code is now more complex but it must be to
work as required.  If you think the are pieces which aren't needed then
let's review those and see if there are simpler techniques.

I too believe that code should be as simple as possible, but first and
foremost it must work.

Larpoux:
There are some words very important for me : 
- Simplicity
- Maintenance
- Continuity

Continuity is very important for our developers. We are not building a new project.
We are setting a new Flutter Sound version.
Every break in the continuity is a a kind of trauma for the user.
And when there is a break, we must be able to explain to the user what is his own benefit.

Changing seconds to milliseconds just because you do not like seconds  is a true case study:
- In Europe we use metric system, and a duration is not measured in hours, minutes or microseconds but seconds.
- But the real thing is that Flutter Sound uses seconds and there is no justification to change that.

We (you) must be able to write a simple notice on how to migrate to this hypothetical 5.0.
If this is not possible because you changed too many things and specially the architecture, then 5.0 will failed its target.

And also I know something very hard to accept : 
A programmer write a bug for every 20 lines he codes. If you rewrite too many things, of course you will fix bugs, but at the same time you will add 2 times more bugs.

bsutton:
Responses in line.

> There are some words very important for me :
> 
> * Simplicity
> * Maintenance
> * Continuity

Agreed.

> 
> Continuity is very important for our developers. We are not building a new project.
> We are setting a new Flutter Sound version.
> Every break in the continuity is a a kind of trauma for the user.
> And when there is a break, we must be able to explain to the user what is his own benefit.
> 
I do agree that continuity is important but there are times when you need to break things in order to move forward.

We also have two user bases that we need to consider.
1) our existing users
2) future users.

The structure of the current code base has grown organically, a little change here, a little change there.
The result is that the api has grown inconsistent and complex.

For example:
Future<String> startRecorder(xxx).

The string used to return the path to a temp file that was created. 
We no longer create a temp file but the method still returns a string.

For a new user the api is hard to understand. This was certainly my experience and I'm not exactly a new to programming.

We also have fundamental problems with the architecture. The current design causes race conditions which is why we are seeing people reporting crashes that we can't find the cause of. 

I've managed to identify and fix most of these problems in the 5.0 release.
To fix these problems I needed to change the way we talk to the OS plugins and this ripples up to how our users need to interact with the api.

The 3.0 api also causes frame drops. This is due to a variety of causes (both in the os plugin and the dart api).  To eliminate these I had to change the way we dealt with futures and these sort of changes ripple all the way up the stack.

If we want flutter sound to be stable in production, these api changes must be made.


> Changing seconds to milliseconds just because you do not like seconds is a true case study:
> 
> * In Europe we use metric system, and a duration is not measured in hours, minutes or microseconds but seconds.
> * But the real thing is that Flutter Sound uses seconds and there is no justification to change that.

Interesting take on SI units. I'm in Australia and we too use the metric system. However I've never heard a developer state the need to use SI when passing time. 
I've changed the streams to return a Duration so the end user can make there own decision and this is more in line with Dart and Java best practices (i.e you shouldn't pass a raw unit of time).

> We (you) must be able to write a simple notice on how to migrate to this hypothetical 5.0.
> If this is not possible because you changed too many things and specially the architecture, then 5.0 will fail its target.

I've updated the read.me on how to migrate and have some more changes coming.
I took me about three minutes to change my existing Player widget to use the new api. 
Of course I understood the details of the api but it will be straight forward change. 
Whilst the api has changed the core concepts have not.

What the change will do is make life easier for our future users and in the end for our existing users.

> 
> And also I know something very hard to accept :
> A programmer writes a bug for every 20 lines he codes. If you rewrite too many things, of course you will fix bugs, but at the same time you will add 2 times more bugs.

We are in complete agreement here.  We need to develop unit test to ensure that the new release works now and into the future.

> Continuity is not only very important for our developers, it is also important for other Flutter Sound coders : Hyo, me, Salvatore, and (we may hope) others.

This a very valid point and I understand it will takes some time for the existing developers to get their head around the changes.

I'm happy to help, answer questions etc and my work is happy to commit resources to fixing bugs in the dart code providing we can get you guys to support the java/ios code.


> If you add a class `Proxy` without any justification, you will just add difficulties to others to understand you and they will loose some times. And time is really what we do not have in excess.

The proxy class has gone now. The original justification was to hide the internal api from our end users.
The 3.0 api exposes a large amount of the internal apis which just serves to confuse users.
After doing some reading I've found a better way to hide the internal apis without the use of a proxy.

As a dev I've always believed that my first job is to make the user's life easy, even if that means I have to work a bit harder.

I've also restructured the code according to googles best practices.
https://dart.dev/tools/pub/package-layout

Note the 'src' directory and the top level flutter_sound.dart class which controls what parts of the api we expose.

I understand you have some significant concerns. 

Can I suggest that we let the users decide.

We can release the 5.0 branch into pub dev as  dev release.

Push out some notices to our users and ask them for feedback.

If they hate the release then we need to look for another path.

bsutton:
I pull this against the wrong branch (again) but I leave this PR here for the moment as it has our discussion thread.

Larpoux:
> Responses in line.
> ...

Globally, I understand you, and I agree with many of your points.

Now, I have a little problem : we have just released a new Flutter Sound version 4.0.0 which includes some minor fixes to some issues.
Please, could you look how to port those fixes to the `dev` branch so that we will not have regression if we eventually publish it ?

bsutton:
Will do.

Larpoux:
I have begun got look to the `dev` branch (Dart side), and there are many things that I like.

I really prefer this code than the previous one I looked to.
I think now that it can be foundations for 5.0 (I was very doubtful until now).

I will take care of the port from 4.0.0 (better that we are not two working on dev), and I think it will not too much hard for me.

I am also going to try to port the 4.0.0 example to dev, so that I will have a clear idea of what a migration to 5.0 means.

bsutton:
> I have begun got look to the `dev` branch (Dart side), and there are many things that I like.
> 
:)
> I really prefer this code than the previous one I looked to.
> I think now that it can be foundations for 5.0 (I was very doubtful until now).
> 
> I will take care of the port from 4.0.0 (better that we are not two working on dev), and I think it will not too much hard for me.

Sound good.
> 
> I am also going to try to port the 4.0.0 example to dev, so that I will have a clear idea of what a migration to 5.0 means.
If you have any questions or suggestions for doco improvements the please let me know.

bsutton:
The RecorderPlaybackController is now working. demo3 shows it in action.

Create a TrackPlayer and a SoundPlayer class and renamed the original SoundPlayer to AudioPlayer.

bsutton:
Closing as was pulled from wrong branch.

PR #327, opened 2020-04-27
Bug fixes and work on being a good citizen

bsutton:
Began work on handling app pause/resume. 
The sound player should now respond correctly however we still need to add plumbing so that the events are sent up from the OS.

I've improved the error paths for the SoundPlayer as the SoundPlayerUI state was being screwed up when errors occurred in SoundPlayer.

Added the method SoundPlayer.rewind

Added new args to both ctors
SoundPlayer.noUI(playInBackground: true);


bsutton:
I note that you have modified a number of old files and pushed them to dev.
Was this intended?
These files are no longer used.
fluauto.dart
track_player.dart

You reverted main.dart however most of this code is now in main_body.dart and has been heavily changed to conform to the new api.
I've reverted main.dart back to the latest version that I've been working on.

bsutton:
I'm not certain why there is a merge conflict on TrackPlayer.m as I've not changed it.

You should take the follow as they have major changes that brings them in line with the new api.

example/lib/drop_downs.dart
example/lib/main.dart
example/lib/main_body.dar

bsutton:
perhaps we could have call where we go through and do a merge?

I'm GMT + 10.
At the time I wrote this it was 3:49 pm.

Larpoux:
> I note that you have modified a number of old files and pushed them to dev.
> Was this intended?
> These files are no longer used.
> flauto.dart
> track_player.dart
> 

It was not intentional. Probably a mistake during my Pull Request integration. I am going to fix that.

> You reverted main.dart however most of this code is now in main_body.dart and has been heavily changed to conform to the new api.

I try to migrate this ancient App to the new  API, so that I will have a clear idea of what it means to migrate.
Also, I do not approve an example/demo App which is more than just a `main.dart` file.

Larpoux:
> perhaps we could have call where we go through and do a merge?
> 
> I'm GMT + 10.

I am GMT +2.
It is now 11:41, and I just get up (I am lazy ;-) ).

bsutton:
migrate.
Also, I do not approve an example/demo App which is more than just a
main.dart file.

Then we probably need to reduce the functionality of the example. Perhaps
we could build several examples that focus on just one aspect of the API

bsutton:
On available until 10pm most nights. What about tomorrow night.

Do you use zoom?

Larpoux:
> Do you use zoom?

What is that ?

bsutton:
A video conferencing software that will allow us to share our desktop.

Larpoux:
I use Skype, FaceTime, Signal (which is very good, free and secured).
I can even use whatsapp
I can use zoom if you think this is better

Larpoux:
I am not very happy with your pull requests.
Too many changes without reasons.
Too complicate without any justification.

When I tried to port `example` to `dev`, I was not able to specify `()onFinished` during my call to `play`... Until I discovered that now it is handled as an attribut.
Why this change ? Do you really want that the user specify his callback AFTER the sound finished ?

You really must understand something : 
nobody never asked for a new flutter sound with more a "sexy" api ("sexy" is just your point of view). A product which will be : 
- more buggy
- incompatible with the previous versions.

What our user ask is : 
- A product more stable and if possible bug free
- The support of other codecs (PCM and OPUS on Android, ...)
- The possibility to play and record with live streams instead of files.

Actually I am considering developing new codecs for Android. But I do not know if I will implement these developments on the current 4.0.x or on your `dev` Pull Request.
Developing new features on the `dev` branch is the only way that our users will want to accept a new version not completely stable. But at the same time, I worry that I will have too much work to do on `dev` for my own taste.

Larpoux:
Having just one class with two different constructor (with or without UI) is dirty.
Because of that, you expose verbs like `startPlayerFromTrack` or the attribut `onSkipToNextTrack` even when NO UI, wich has no sense.

My own appreciation is that having two classes, with one inheriting from the other was clean.

Larpoux:
I have not yet understood how to manage playback and recording subscriptions.
Probably I will have to "Read the Fucking Doc ;-) .
But for me it is very often simpler to read code than to read English.
And the code is always accurate !

Larpoux:
> Do you use zoom?

I installed Zoom on my computer. My account is larpoux@gmail.com. But Zoom is again another software neither free nor open source. And no end-to-end encryption. Not better than Skype or FaceTime, I think. "Signal" is FOSS and has end-to-end encryption. But OK for Zoom if you prefer...

Larpoux:
The reason I separated TrackPlayer in another class, using another channel was because "TrackPlayer" was new code, and I did not trust Salvatore's code.
I really did not want to have regression with "FlutterSoundPlayer".

Now things are a little different. We know that TrackPlayer is very buggy, but not too much.
Probably not a great deal to merge the two classes. But why ? Having split the code in two separated blocs was not very easy (Salvatore mixed everything together) but has major benefits. The result is clean, both for our developers and for us. I do not understand what would be the advantage to remix everything together and get a dirty code.

bsutton:
> > Do you use zoom?
> 
> I installed Zoom on my computer. My account is [larpoux@gmail.com](mailto:larpoux@gmail.com). But Zoom is again another software neither free nor open source. And no end-to-end encryption. Not better than Skype or FaceTime, I think. "Signal" is FOSS and has end-to-end encryption. But OK for Zoom if you prefer...

I've installed signal. How do we connect to each other?

bsutton:
> I have not yet understood how to manage playback and recording subscriptions.
> Probably I will have to "Read the Fucking Doc" ;-) .
> But for me it is very often simpler and easier to read code than to read English.
> And the code is always accurate !

The management is now all done in a single call.

Stream<PlaybackDisposition> stream = SoundPlayer.noUI().dispositionStream(Duration interval);
Stream<RecordingDisposition> stream = SoundRecorder.noUI().dispositionStream(Duration interval);

If you are building a UI then using a StreamBuilder is the easiest way to handle the streams.

Larpoux:
I installed zoom. How to join you ?

Larpoux:
My ID on zoom is larpoux@gmail.com

bsutton:
> Having just one class with two different constructors (with or without UI) is dirty.
> Because of that, you expose verbs like `startPlayerFromTrack` or the attribut `onSkipToNextTrack` even when NO UI, which has no sense.
> 
> My own appreciation is that having two classes, with one inheriting from the other was clean.

This decision was not a simple one and I tried variations (including the original concept of two separate classes).

In the end I settled on a single class for a couple of reasons.

The push these details is to minimize the use of inheritance as it as complexity to the code base.
The recommendation these days is to use encapsulation. 
I have used some inheritence in the code base but I've tried to keep it to a minimum.
Certainly when I started looking at the 3.0 code the inheritance the TrackPlayer used made it harder to understand. Which was certainly one of the drives to move away from inheritance for this class.

I also don't completely like the fact that we expose the skipforward, skipbackward methods because as you correctly point out they current don't do anything when using the .noUI() constructor.

A few things made me leave them as is.
1) my desire to avoid inheritance for this class
2) The method level documentation should be enough to flag that they don't do anything.
3) and this was the most important one. I'm considering making them usable for the .noUI constructor. The idea being that a developer may want to build on our code and inject skip events through to a third party UI. 

So in the end it was a trade off.  Its not terrible (nor perfect) and as yet it may be useful.

I'm confused about your statement:
> You expose verbs startPlayerFromTrack

The SoundPlayer class does not expose this method.
The SoundPlayer exposes a single method:

```play(Track)```

This method works the same way whether you have a UI or not.

Except for the skip properties all other methods work identically regardless of the constructor you use.

**Comment by Larpoux** [comment]
:-D  You should work with Salvatore, and not me :-D

Salvatore wanted just one class with a flag specifying if the user wants withUI or witoutUI.
I did not agree, and I isolated the TrackPlayer from the rest of flutter sound.

**Comment by Larpoux** [comment]
`play(Track)` to play a "bip" or "click" is not very judicious, I think.

**Comment by bsutton** [comment]
> When I tried to port `example` to `dev`, I was not able to specify `()onFinished` during my call to `play`... Until I discovered that now it is handled as an attribut.
> Why this change ? Do you really want that the user specify his callback AFTER the sound finished ?

The reason for this change was to solve forward compatibility problems.
I was concerned that if we put the callbacks in the constructor over time we would end up with a constructor with a large no. of arguments that most people are not interested in.

By moving the callbacks into properties we can continue adding callbacks as needed without polluting the constructor.

SoundPlayer allows you to set the callbacks BEFORE you start playing e.g.

```dart
var player = SoundPlayer.noUI();
player.onFinished = () => player.release();
player.onStop = () => player.release();
player.play(track);
```
The above code guarantees that the sound player will be released.

The second advantage of make the onFinished a property is that if necessary you can change what the do after creating the player.  Although I really don't view this as a core driver of the design.

I've provided extensive examples in the readme.
I plan on providing examples in the api documents so you will be able to see these examples when doing auto complete in your ide.

Larpoux:
OK Brett, I am going to sleep, now. Talk with you tomorrow.

bsutton:
> You really must understand something :
> nobody never asked for a new flutter sound with more a "sexy" api ("sexy" is just your point of view). A product which will be :
> 
> * more buggy
> * incompatible with the previous versions.
> 
> What our user ask is :
> 
> * A product more stable and if possible bug free
> * The support of other codecs (PCM and OPUS on Android, ...)
> * The possibility to play and record with live streams instead of files.

I think you have misunderstood my motivation for being involved in this project so perhaps I should take a moment and try to explain why I'm here.

Lets start by taking about code maintenance.

You have raised concerns that nobody is helping you maintain the 3.x branch.

First I completely agree that maintenance comes before features.
However when involved with a project I always try to use the best person for the job.

I've intentionally not touch the 3.x branch as I don't understand it sufficiently to make significant changes. 
Given you familiarity with the 3.x code base you are much better positioned to do the maintenance on the 3.x branch.

I fully expect to take on most of the maintenance work for the 5.x branch given I will be the one most familiar with it.

Here is how I thought things might work.

State 1)
I work on developing the 5.0 branch which has the following goals
* provide a simple api
* provide a consistent api
* architectural changes to ensure that we can eliminate the race conditions in the existing 3.x code base.
* Improve documentation to help new users get started with Flutter Sound.

During stage 1 you continue to do maintenance fixes on 3.0 and make the small no. of changes in the java/object-c code required for 5.0 (this is just the onConnect event).

Stage 2) (date to be determined) Post release of 5.0
I provide maintenance support for the 5.0 dart code.
You provide maintenance support for the 5.0 java/objective-c code

Stage 3) additional development on the 5.0/6.0 +  branch
We work out what needs to be done and the person with the most experience/time in that area does the dev.

Hopefully the additional level of documentation in the code will encourage other developers to get involved and we can grow the feature set.

It might help if I explain my background.
I've have no knowledge of audio systems however I have a background in realtime event processing system from developing Stock Broking systems and more recently Telephony systems for contact centres. I'm a minor contributor to Asterisk and the maintainer of AsteriskJava (https://github.com/asterisk-java/asterisk-java) and the dart library Money2 (https://pub.dev/packages/money2).

Due to the nature of these system I have a large amount of experience in multi-threaded applications and event processing models.  Race conditions is something that I have to deal with on a daily basis.

I'm fairly new to flutter (about 12 months). I still struggle a bit with the widget model (communicating data between widgets) but I have a fairly strong grasp of dart (I have a hobby project https://pub.dev/packages/dshell which allows you to replace bash with dart).

My coding background is C/C++/Java and now dart. 

So that's my bio. It might help if I knew a little more about your background and experience. In this way we can each understand where the others strengths are which will help us divide up the project.

Now let me explain my motivation for being involved in this project.

The company that I work for is doing a major flutter development which requires the ability to record and play back audio.

We are using flutter sound to do this.
Our code is already working with the 3.x code branch but it has been plagued with crashes.
We eventually worked at that it was due to a no. of race conditions in the 3.x code base.

So my company needs the sound system to be stable.
Given the importance of the sound system we decided to hop in and help with the development which is where my involvement comes from.

We are prepared to put in a long term commitment into helping maintain the code base.

bsutton:
Just to keep you up to date. I'm currently working on migrating a recording widget into the code base.

Once done I will look a creating unit tests for the api.

bsutton:
> `play(Track)` to play a "bip" or "click" is not very judicious, I think.

That's what the QuickPlay class is for.

bsutton:
> The reason I separated TrackPlayer in another class, using another channel was because "TrackPlayer" was new code, and I did not trust Salvatore's code.
> I really did not want to have regression with "FlutterSoundPlayer".
> 
> Now things are a little different. We know that TrackPlayer is very buggy, but not too much.
> Probably not a great deal to merge the two classes. But why ? Having split the code in two separated blocs was not very easy (Salvatore mixed everything together) but has major benefits. The result is clean, both for our developers and for us. I do not understand what would be the advantage to remix everything together and get a dirty code.

I'm not actually suggesting that you mix the two classes within the java code. Looking at the code they are quite different. The code could do with a re-organisation as its fairly hard to understand but I'm all for small classes which do a single job. 

I was suggesting a single plugin architecture as it simplifies the communications between dart and java.
I wasn't commenting on how to structure the classes within the plugin.

bsutton:
lets zoom or signal when you get up.

Larpoux:
> I fully expect to take on most of the maintenance work for the 5.x branch given I will be the one most familiar with it.

👍  Great! Very great. I would not like to maintain 5.0 and when bugs you would tell me 'This bug is probably fixed in 6.0'

> Hopefully the additional level of documentation in the code will encourage other developers to get involved and we can grow the feature set.

This is certainly what I appreciate with your job, and want to keep that in the released version.
My wish is to have a perfect score on "pub.dev". Actually Google complains that our plugin does not respect the camelCase recommendations and does not have comment on public procedures.
I gess that adding all those comments was not really a funny job, and I thank you for having done it.

> Due to the nature of these system I have a large amount of experience in multi-threaded applications and event processing models. Race conditions is something that I have to deal with on a daily basis.

I hate Dart/Flutter `Future` and `async` . Probably more than half the bugs I have in my app are because of `async` processing.
There is just one thing that I hate more than the Dart async : threads.
Threads are just bug nests. And a correct maintenance of a program which use threads is almost impossible. Fortunately, Dart/Flutter does not use threads.

> It might help if I knew a little more about your background and experience.

I began my programmer life when I was 18 (I am now almost 70).
So I have more than 50 years experience in coding, and I am not going to send you my resume.
I have worked with many languages, mostly Pascal which is a language that I love, C++ and Java.
I had to manage teams as a project leader but I do not like that, and I am not good as a `boss/chief`. I really prefer to code, which is something that I love.
During many years I had to maintain code written by others.
I understand now this : 
- Coding is very simple and easy. Everybody can code, specially code things that do not work.
- Coding code which works is more difficult
- But getting working code written by others and did not work is the real challenge. This is called maintenance. Maintenance is the real Art of the developer. It requests very various qualities that we do not find on many people.
I do not pretend to have all the required qualities, but I do my best.
Of course, maintenance is not funny, and I would prefer to code.

I am now retired and I continue to code because this is the only thing that I am able to do in my life.

Larpoux:
> > `play(Track)` to play a "bip" or "click" is not very judicious, I think.
> 
> That's what the QuickPlay class is for.

This is not possible to create and open a new Audio Session each time the App wants to play a "Bip". You will have many problems, specially with the Audio Focus.

bsutton:
So how do you play a short sound?

What are the alternatives?

bsutton:
I've sent you a zoom invite.

I met Richard once, an interesting character.

bsutton:
@Larpoux 
Thanks for taking the time to chat last night.

Just a couple of notes in writing in case some things got lost in translation.

Track
I'm now using Tracks everywhere for both version of SoundPlayer.withUI/noUI and for the SoundRecorder.

You mentioned that the track details such as album art can't be used when calling the noUI constructor.
The SoundPlayerUI widget I've built actually already uses track details (artist and album) and at a future date I would like to support album art. 
So using a track for both players provides a consistent api and is being used by both versions of SoundPlayer.

I've also changed to using a Track for the SoundRecorder.
There are a couple of reasons.
1) consistency. You can now recording into a Track and then play that same track back.
So you can now call:
```dart
var track = Track.fromPath(...)
SoundRecorder().record(track);
...
SoundPlayer.withUI().play(track);
```
2) the second reason is that I'm writting a widget that takes advantage of the above feature.

```dart
Widget build(BuildContext context)
{
  return  RecorderPlaybackController(
       child: Column(
          children: [
            SoundRecorderUI(track),
            SoundPlayerUI.fromTrack(track),
          ],
        ))
```
You will note in the above code that the same track is passed to the recorder and the player.
The RecorderPlaybackController synchronises the two widgets so that you can record and then playback the same track.
The controller won't allow you to play whilst recording and updates the players duration as the recording proceeds.

So passing a Track to both recorder and player creates way of passing audio around a system.

# The two classes discussion
I think the biggest remaining issues is the two constructors for SoundPlayer.

You would like to split this into two separate classes, I would prefer to leave this as one class with two constructors.

So here is my argument for leaving it as is.

1) when I started using Flutter Sound it was not at all clear what the difference was between the TrackPlayer and the SoundPlayer. I had to dig into the code/doco to work this out.
The two constructors I've used give a better indicator of the difference and the appropriate use case.

2) the api for both is now identical
With the SoundPlayer all of the api is available for both withUI and noUI.
As explained, above I use Tracks for both constructors.
As for the skip callbacks I have plans to use these in the noUI version and third parties may also want to use them as they provide a consistent api between the two constructors.

At this point the entire api for both constructors is consistent and meaningful.
If we split them we would end up with two classes with identical api just a different name.

3) Splitting the SoundPlayer into two classes will actually make the code more complicated.
Since the SoundPlayer class is fully functional and the logic is quite simple we probably don't want to split its code out into two classes.
If we did we would be duplicating significant amounts of boiler plate.

If we want to avoid the boiler plate we have two options.
Inheritance or encapsulation.
If we use encapsulation we would have to proxy all of the public api calls through to the SoundPlayer class. Hence lots of boiler plate we have to maintain.

If we use inheritance we can add two fairly thin classes on top of the SoundPlayer but what do we get for it.
The user sees two public classes rather than two public constructors.

I don't see any material difference in the two approaches and we have just created more code for no tangible gain.

# Unit Testing
I would really like to see us spend some time creating unit tests.
In my other projects where I've used this it saves testing time on each release and delivers a higher quality product with less bugs.

# Codec support.
We spoke briefly about codec support.

I would like to propose that we keep the core Flutter Sound package as simple as possible.
The core package could even just support a single codec that is support natively on both iOS and Android.

We then modify the SoundRecorder and SoundPlayer apis to allow a codec converter to be plugged in.

Perhaps something like this.
```dart
/// register a new codec by name.  
Codec newCodec = Codec.register({String name, String extension});

/// Register a codec converter
CodecConverter.register( CodecConverter converter);

abstract class CodecConverter
{
  static  List<CodecConverter> registered = [];

   CodecConverter({Codec fromCodec, Codec toCodec});
  
  static void register(CodecConverter converter)
  {
    registered.add(converter);
  }
}

class CodecConverterFile
{
   convert({String fromPath, String toPath});
}

class CodecConverterBuffer
{
      convert({Uint8List fromBuffer, Uint8List toBuffer});
}
```
We would need to change the Codec enum to be a class with string constants but that's easy to do.

We can then ship an ancillary package - Flutter Sound Codecs that contains a growing list of codecs.

This would help keep the core Flutter Sound package simple which would make testing easier. This will have the side affect that we can ship the 5.0 release sooner and get people using the core libraries in beta whilst we prepare the codecs.

This method would also allow third parties to add their own codec support.

Thoughts?

Larpoux:
Hi Brett,
Could you look to this patch, please ?
[0001-Working-on-the-old-example-app.patch.zip](https://github.com/dooboolab/flutter_sound/files/4550784/0001-Working-on-the-old-example-app.patch.zip)

Larpoux:
@bsutton : could you look to this patch, please.
I had to remove `_closeDispositionStream()` from `stop()`
```
// Stops playback.
  Future<void> stop() async {
    initialised.then<void>((result) async {
      if (result == true) {
        if (isStopped) {
          Log.d("stop() was called when the player wasn't playing. Ignored");
        } else {
          try {
            // playerController is closed by this function
            //_closeDispositionStream();
            await _plugin.stop(this);
            playerState = PlayerState.isStopped;
            if (_onStopped != null) _onStopped(wasUser: false);
          } on Object catch (e) {
            Log.d(e.toString());
            rethrow;
          }
        }
      } else {
        throw PlayerInvalidStateException('Player initialisation failed');
      }
    });
  }
```
and an API verb `initialize()` in `SoundRecorder`
```

  /// Initialize a fresh new SoundRecorder
  Future <SoundRecorder> initialize()  =>_initialize();
```
[0001-Working-on-the-old-example-app.patch.zip](https://github.com/dooboolab/flutter_sound/files/4550860/0001-Working-on-the-old-example-app.patch.zip)

bsutton:
I've actually done quite a bit of work on the internals do it's changed
quite a bit now. Will push an update tomorrow and we can merger then.

Larpoux:
The code with async/future is incorrect (which is not surprising for me 👎 ).
When the app does : 
` await playerModule.stop();`
it get sometimes a return before the player is really stopped.

Larpoux:
Same problem with : 
```
if (playerModule != null) await playerModule.release();
```
which gets a return before the playerModule is really released.
I think that this problem occurs with all the API verbs.

I hate Futures and Async procedures !

Larpoux:
Hi Brett,
here my actual desiderata : 

1. An example(demo) with just one main.dart

2. A clear separation of the playback API in two classes
    - SoundPlayer
    - TrackPlayer

3. Constructors which does only variable initialization. A constructor should not do processing and should not be an async procedure nor return a future (of course)

4. Two API verbs : `openAudioSession(...)` and `closeSession()`

5. `openSession` will have the following parameters : 
     - audioFocus:
          - ACQUIRE_AUDIO_FOCUS_AND_STOP_OTHERS
          - ACQUIRE_AUDIO_FOCUS_AND_DUCK_OTHERS
          - ACQUIRE_AUDIO_FOCUS_AND_KEEP_OTHERS
          - DO_NOT_ACQUIRE_AUDIO_FOCUS

     - onPause:  (just for TrackPlayer)
          - HIDDEN
          - DISABLED
          - myCallbackFunction
          - AUTO
            (when auto, the pause button will be disabled if not playing, or will pause or resume depending on the player state)

     - onSkipForward: (just for TrackPlayer)
          - HIDDEN
          - DISABLED
          - myCallbackFunction

     - onSkipBackward: (just for TrackPlayer)
          - HIDDEN
          - DISABLED
          - myCallBackFunction

     - showPauseButton: (just for TrackPlayer)
          - HIDDEN
          - DISABLED
          - PAUSED
          - PLAYING
          - AUTO

6. audioFocus(...). Will have the following parameter : 
     - ACQUIRE_AUDIO_FOCUS_AND_STOP_OTHERS
      - ACQUIRE_AUDIO_FOCUS_AND_DUCK_OTHERS
     - ACQUIRE_AUDIO_FOCUS_AND_KEEP_OTHERS
     - ABANDON_AUDIO_FOCUS

The audio focus must not be changed by flutter_sound internal during `startPlayer()` and `stopPlayer`. It will be acquired during `openSession()` if specified in the parameter, and abandoned during `closeSession()`

7. play(...)
     - uri: The URI of the sound to be played or track: (for TrackPlayer)
     - Codec:
     - onFinished: an eventual call-back function.

8. showPauseButton(...).  (just for TrackPlayer). Will have the following parameter : 
     - HIDDEN
     - DISABLED
     - PAUSED
     - PLAYING
     - AUTO

9. setElapsedTime(...)

Note: Dyglot Korean App must be able to specify dynamically the icon state for the pause button, and the playback to call, **even when no sound is currently played back**.

Note: Dyglot Korean App must be able to specify dynamically the progress bar on the lock screen, **even when no sound is currently played back**

bsutton:
This is not a problem with the async code its actually a problem with the
java/objective-c code.

The java code calls result.success before the media actually stops.

What I would like to see is that when we call into the java code we start
the required action and call result.success.

When the action completes  the java code calls invokeMethod to with a
success/failure status e.g. onStopped(true/false).
This is what I did for the initialisation logic which now sends back an
onConnected call from the java.

We need this for initialise, play and stop.
I don't think the other methods are a problem at this point.

If you are happy to write the java/objective-c code I can mode the dart
code to handle the callbacks.

bsutton:
I doubt that this is really a problem.

Once the release is called the SoundPlayer is shutdown.

When we start another SoundPlayer instance it correctly waits for the
initMediaPlayer call to complete which would in turn wait on any prior
release.

Unless you have found a scenario that I'm not aware of?

bsutton:
Questions in line:


> Hi Brett,
> here my actual desiderata :
> 
> 1. An example(demo) with just one main.dart
What is your objective here?

I think the current example is too complex its more of a demonstration of capabilites.

I would be useful to have a no. of actual examples that just show very simple use cases. In fact the flutter guidelines recommend this.

I would suggest renaming the existing example to be 'demo.dart'.
I would then suggest we create a number of small examples.
1) player UI
2) recorder UI

We then can create example code in the doco on how to use the api. I've already created a fair amount of this.
Google recommends adding examples into the doco for each method. I've done some of this.


> 2. A clear separation of the playback API in two classes
>    
>    * SoundPlayer
>    * TrackPlayer
I've renamed the existing SoundPlayer to AudioPlayer and I will create to classes SoundPlayer and TrackPlayer.


> 3. Constructors which does only variable initialization. A constructor should not do processing and should not be an async procedure nor return a future (of course)
What is the technical reason behind this?
As noted dart doesn't allow async  ( of course :).

This used to be a problem in the early C++ days but I'm not aware of any technical nor design pattern that suggests this is a problem with Garbage Collected languages.

> 4. Two API verbs : `openAudioSession(...)` and `closeSession()`
>    `openSession` will have an audio Focus parameter :

Should this argument not be on the play method?
If its on an openSession method then you can change this between plays?
>    
>    * ACQUIRE_AUDIO_FOCUS_AND_STOP_OTHERS
>    * ACQUIRE_AUDIO_FOCUS_AND_DUCK_OTHERS
>    * ACQUIRE_AUDIO_FOCUS_AND_KEEP_OTHERS
>    * DO_NOT_ACQUIRE_AUDIO_FOCUS

> 5. audioFocus(...). Will have the following parameter :
>    
>    * ACQUIRE_AUDIO_FOCUS_AND_STOP_OTHERS
>    * ACQUIRE_AUDIO_FOCUS_AND_DUCK_OTHERS
>    * ACQUIRE_AUDIO_FOCUS_AND_KEEP_OTHERS
>    * ABANDON_AUDIO_FOCUS
> 
I would prefer a separate 'abandonFocus' rather than a enum of 'ABANDON_AUID_FOCUS'

> The audio focus must not be changed by flutter_sound itself during `startPlayer()` and `stopPlayer`. It will be acquired during `openSession()` if specified in the parameter, and abandoned during `closeSession()`

As above why not allow it on a per play basis?

> 
> 1. onLoosingFocus(...)
Sounds good.
>    
>    * myCallbackFunction
>    * PAUSE
>    * STOP
>    * DUCK
>    * KEEP
I don't understand what the above enum is about?

> 2. showPauseButton(...).  (just for TrackPlayer). Will have the following parameter :
>    
>    * HIDDEN
>    * DISABLED
>    * PAUSED
>    * PLAYING
>    * AUTO

Can you explain these enums?

> 3. onPause:  (just for TrackPlayer)
>    
>    * myCallbackFunction
>    * AUTO
>      (when auto, the pause button will be disabled if not playing, or will pause or resume depending on the player state)

> 4. onSkipBackward(...) (just for TrackPlayer)
>    
>    * HIDDEN
>    * DISABLED
>    * myCallBackFunction
> 5. onSkipForward() (just for TrackPlayer)
>    
>    * HIDDEN
>    * DISABLED
>    * myCallbackFunction
> 6. setElapsedTime(...)

What does this do?

>    
>    * Value
>    * AUTO

> 7. play(...)
>    
>    * uri: The URI of the sound to be played or track: (for TrackPlayer)
>    * Codec:
>    * whenFinished: an eventual call-back function.
> 8. API verbs must be functions
>    Example:
>    
>    * isPlaying()
>    * getAudioState()
This actually goes against the dart guidelines:

From the guidelines:
https://dart.dev/guides/language/effective-dart/design
'When a member produces a result without any side effects, it should usually be a getter '
`AVOID starting a method name with get.`

So isPlaying
bool get isPlaying => player.state == playing;
audioState
PlayerState get audioState => player.state;


> 9. Nouns must be attributs or instance variables
>    Example:
>    
>    * audioState
> 
> Note: Dyglot Korean App must be able to specify dynamically the icon state for the pause button **even when no sound is currently played back**.
> 
> Note: Dyglot Korean App must be able to specify dynamically the progress bar on the lock screen, **even when no sound is currently played back**
> 
> This is my own point of view of a nice API.
> We should talk together to find a good compromise/trade-off with your own point of view

bsutton:
> @bsutton : could you look to this patch, please.
> I had to remove `_closeDispositionStream()` from `stop()`
> 
> ```
> // Stops playback.
>   Future<void> stop() async {
>     initialised.then<void>((result) async {
>       if (result == true) {
>         if (isStopped) {
>           Log.d("stop() was called when the player wasn't playing. Ignored");
>         } else {
>           try {
>             // playerController is closed by this function
>             //_closeDispositionStream();
>             await _plugin.stop(this);
>             playerState = PlayerState.isStopped;
>             if (_onStopped != null) _onStopped(wasUser: false);
>           } on Object catch (e) {
>             Log.d(e.toString());
>             rethrow;
>           }
>         }
>       } else {
>         throw PlayerInvalidStateException('Player initialisation failed');
>       }
>     });
>   }
> ```
> 
> and an API verb `initialize()` in `SoundRecorder`
> 
> ```
> 
>   /// Initialize a fresh new SoundRecorder
>   Future <SoundRecorder> initialize()  =>_initialize();
> ```
> 
> [0001-Working-on-the-old-example-app.patch.zip](https://github.com/dooboolab/flutter_sound/files/4550860/0001-Working-on-the-old-example-app.patch.zip)

I've fixed the issue with stop you noted above.

I still do not understand the need for a public initialise method.
It adds no additional functionality.
It makes the api (slightly) harder to use.
There is no technical nor design pattern that I'm aware of the requires/recommends it.

bsutton:
I need some help with the audio focus.
1) I don't know how to stop others
2) have I don the duck correctly?

This is the code so far.
```duck
  ///  The caller can manage his audio focus with this function
  /// Depending on your configuration this will either make
  /// this player the loudest stream or it will silence all other stream.
  Future<void> audioFocus(AudioFocusMode mode) async {
    initialised.then<void>((result) async {
      if (result == true) {
        switch (mode) {
          case AudioFocusMode.focusAndKeepOthers:
            await _plugin.audioFocus(this, request: true);
            _setHush(hushOthers: false);
            break;
          case AudioFocusMode.focusAndStopOthers:
            await _plugin.audioFocus(this, request: true);
            // TODO: how do you stop other players?
            break;
          case AudioFocusMode.focusAndDuckOthers:
            await _plugin.audioFocus(this, request: true);
            _setHush(hushOthers: true);
            break;
          case AudioFocusMode.abandonFocus:
            await _plugin.audioFocus(this, request: false);
            break;
        }
      } else {
        throw PlayerInvalidStateException('Player initialisation failed');
      }
    });
  }

  /// Apply/Remoe the hush other setting.
  void _setHush({bool hushOthers}) async {
    if (hushOthers) {
      if (Platform.isIOS) {
        await iosSetCategory(
            IOSSessionCategory.playAndRecord,
            IOSSessionMode.defaultMode,
            IOSSessionCategoryOption.iosDuckOthers |
                IOSSessionCategoryOption.iosDefaultToSpeaker);
      } else if (Platform.isAndroid) {
        await _androidFocusRequest(AndroidAudioFocusGain.transientMayDuck);
      }
    } else {
      if (Platform.isIOS) {
        await iosSetCategory(
            IOSSessionCategory.playAndRecord,
            IOSSessionMode.defaultMode,
            IOSSessionCategoryOption.iosDefaultToSpeaker);
      } else if (Platform.isAndroid) {
        await _androidFocusRequest(AndroidAudioFocusGain.defaultGain);
      }
    }
  }
```

bsutton:
closed by accident.

bsutton:
I'm not a fan of the whenFinished method as its inconsistent with the rest of the api.

We have a large and growing no. of callbacks so it doesn't work to add all the callbacks to the constructor/initialise nor play methods.

The whenFinished method is in no way special.
I think originally you had a concern that the whenFinish had to be on the constructor otherwise it wouldn't be called.
With the current design that isn't the case.

```dart
    var _player  = SoundPlayer();
    _player.setVolume(volume);
    _player.audioFocus(AudioFocusMode.focusAndDuckOthers);
    _player.onFinished = () => _player.release();
    _player.onStopped = ({wasUser}) => _player.release();
    return _player.play(_track);
```

The above code guarentees that onFinished will be called (assuming the playback completes naturally).

It also means that onFinished is set in the same way that onStopped is set and the close to a dozen other callbacks are all set in the same way.

bsutton:
Adding a callback `onLoosingFocus` seems like a good idea however I can't see any java code that emits the required event:

The closest I can see is in BackgroundAudioService.java:

```

	@Override
	public void onAudioFocusChange( int focusChange )
	{
		// The audio focus has changed, then take an action based on the new audio focus
		switch ( focusChange )
		{
			case AudioManager.AUDIOFOCUS_LOSS:
			{
				// Another app has requested audio focus, then stop audio playback
			}
			case AudioManager.AUDIOFOCUS_LOSS_TRANSIENT:
			{
				// Another app wants to play audio for a short time, then pause audio playback
				mMediaSessionCallback.onPause();
				break;
			}
			case AudioManager.AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK:
			{
				// Another app requested focus, but we can continue the media playback
				// bringing the volume down a bit.
				if ( mMediaPlayer != null )
				{
					mMediaPlayer.setVolume( 0.3f, 0.3f );
				}
				break;
			}
			case AudioManager.AUDIOFOCUS_GAIN:
			{
				// The audio focus was granted: either a request was accepted, or a
				// AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK event has completed.
				if ( mMediaPlayer != null )
				{
					// Restart the player if it was not playing
					if ( !mMediaPlayer.isPlaying() )
					{
						startPlayerPlayback();
					}
					// Set the volume to its previous levels
					mMediaPlayer.setVolume( 1.0f, 1.0f );
				}
				break;
			}
		}
	}
```
If you can flesh this out to send the events then I can write the dart code.

bsutton:
I've fleshed out TrackPlayer and SoundPlayer classes using encapsulation to avoid duplicating code.

As you want to reduce the api inheritance isn't an option..

The underlying api is now encapsulated in AudioPlayer. 

I don't understand all your requirement or the logic behind some of them so will need further discussion.
Some of the requirements would also appear to need changes to the java/objective-c code.

bsutton:
You mention that you only want onPause for track player.

I think this is wrong as if a third part app starts playing audio and asks the OS to stop us playing then I believe that we can get both 'pause' and 'stop' events coming up from the OS. 
If this is the case then the SoundPlayer needs to be able to emmit both of these events.

bsutton:
I've just being doing some testing and it looks like the onFinished doesn't work quite like I expect it to.

The doco states:
`Register a callback to be invoked when the end of a media source has been reached during playback.`

I've interpreted this to mean we get a onFinished once the audio naturally reaches its end.

However if I stop the media playing early I'm also seeing an onFinished event come up from the OS.

As such it appears we can't easily differentiate between natural completion and a forced stop.

As such I'm thinking that we should merge the onFinished and onStopped callbacks into one callback.

I would suggest that it should be onStopped as (in english) this more accurately reflects what is happening.

Larpoux:
Why `AudioPlayer` and `SoundRecorder` ?
This is not homogeneous
- Better `AudioPlayer` and `AudioRecorder`
- or `SoundPlayer` and `Soundrecorder`

bsutton:
Audio player is the implementation class for soundplayer and trackplayer

PR #333, opened 2020-05-01
Made the storage type explicit to simplify the logic. You can now pas…

bsutton:
You can now pass a null databuffer to fromBuffer which is useful if you a recording.


bsutton:
Bug fixes, performance improvements and usability.

Larpoux:
Hi Brett,

I plan to release a new Flutter Sound version sometime next week.
This new version will have a few breaking changes. Not many, fortunately.
This version will support new codecs for Playback or recording, both on Android and iOS.
This version will support PCM recording, which is something that many of our customers are waiting for.
It will have also a new API verb to record PCM to a Dart Stream, instead of a file.

If you want to include some Pull Request to the Master Branch, it is the good moment.
What I would like is not a Pull Request of the "dev" branch but several Pull Request, each small enough to be able to respect continuity and check that there will not be any regression.

bsutton:
> If you want to include some Pull Request to the Master Branch, it is the good moment.
> What I would like is not a Pull Request of the "dev" branch but several Pull Request, each small enough to be able to respect continuity and check that there will not be any regression.

This is probably not practical. 
The dart code is significantly different that small pulls would require re-writes.
The java code could possibly be done as a pull as the only significant change is that it sends a 'onConnected' event at the end of initMediaPlayer which could just be ignored.

I'm not certain that it is worth the effort at this point.
I'm inclined to get the 4.0 release out and get the 5.0 into its own beta so we can get feedback from users on the api changes.

Larpoux:
What I would like to do, is to release a V5 with not too much API changes, but with all those which will be used in the foreseeable future. Even if not actually necessary.
I hope that the V5 API will be stable enough so that we can continue to enhance Flutter Sound without having to release a V6 in a few weeks. Our users will not be happy if we introduce breaking changes every months.

bsutton:
The reason I'm pushing to get the 5.0 beta out is so we can get some feed back.
I was thinking that we have a longer than usual beta period given the scope of the changes.

The principle that I was working on with 5.0 was that if we are going to break the api then lets break it all at once.
I'm reasonably confident that the api will hold up over time but still want to expose it to the community for comment.

I really don't think that we can release 'parts' of the v5 api. Its a major refactoring of the code base so its not going to be practical to back port pieces of it.

I would suggests that we get 4.0 out as the 'stable' branch and plan on 4-6 weeks of v5 in beta.

I'm currently implementing it into our production system. If you are able to do the same into your korean app then we would have a couple of in development systems that we can test and fix bugs against. (I'm not certain where you are in your release cycle to know if this is practical).

So we have another conference call to night to go over some of these details and finalise a few of the remaining 5.0 api issues?

Larpoux:
> I really don't think that we can release 'parts' of the v5 api. Its a major refactoring of the code base so its not going to be practical to back port pieces of it.

This is really the point that I do not validate. We *must* maintain and enhance Flutter Sound with incremental changes. If we rewrite everything, this is a new project. Not Flutter Sound any more. If we try to revolution everything in Flutter Sound, it will be the best way to kill it.

bsutton:
The counter argument is that if you keep making incremental changes you are really going to annoy users as they will have to keep updating their code.

The aim of 5.0 is to make one large set of changes that should fix everything at once.
The aim of the 5.0 api design was to stop the ongoing incremental changes and fix the large architectural issues.

We are in the very early days of flutter with more developers coming to joining the flutter community on a daily basis.

Whilst we must look after our existing users we also need to be attractive to new users. 
The current api is less than ideal and the constant tweaks we will have to do are just going to get annoying. The api also had to be change in some fundamental ways (futures). Little tweaks are not going to fix these.

I'm proposing the extended beta so we give users a chance to review and comment on the api so we can get it right before we do a full release.

If we break the api once, we will be forgiven. If we break the api every other week we will be hated.

bsutton:
App Pause/Resume not correctly frees media resources.
The SoundPlayerUI will pause and auto rewind 1 second when the app is resumed.

bsutton:
The SoundRecorder now correctly frees resources when we see app pause/resume events.

PR #339, opened 2020-05-05
Fix Small Typing Issues in Docs

abegehr:
- use await instead of Future+await
- remove then(), since using await
- startRecorder() expects uri as optimal parameter for filePath


Larpoux:
Hi @abegehr .
Thank you for your Pull Request.
I am currently rewriting the documentation. During the next few days, this documentation will be unstable.

I realize that `Future`, `await` and `then:` are actually incorrect in the documentation.
I am going to fix that.

Thank you again 👍

Issue #340, opened 2020-05-05
NullPointerException at fluttersound (initMediaSession.BackgroundAudioService.java:565) [BUG]


3mrdev:
## Flutter Sound Version : 

- **Released** or **Beta** version ?
Released

- Version number ?
^4.0.3+1

- **FULL** or **LITE** flavor ?
LITE i guess .. to support android api 16
flutterFFmpegPackage  = "audio-lts"

----------------------------------------------------------

## Severity

- Crash 

--------------------------------------------------------

## Platforms you faced the error 

- Android 
- Real device 

----------------------------------------------

## Logs
(This is very important. Most of the time we cannot do anything if we do not have information on your bug)

Google Play crash report ( i have more for other phones request them if it will help you i will be happy to help)

34 minutes ago on app version 15
Huawei Y7 Prime 2019 (HWDUB-Q), 3072MB RAM, Android 8.1

```
java.lang.RuntimeException: 
  at android.app.ActivityThread.handleCreateService (ActivityThread.java:3976)
  at android.app.ActivityThread.-wrap5 (Unknown Source)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2109)
  at android.os.Handler.dispatchMessage (Handler.java:109)
  at android.os.Looper.loop (Looper.java:166)
  at android.app.ActivityThread.main (ActivityThread.java:7555)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:469)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:963)
Caused by: java.lang.NullPointerException: 
  at com.dooboolab.fluttersound.BackgroundAudioService.initMediaSession (BackgroundAudioService.java:565)
  at com.dooboolab.fluttersound.BackgroundAudioService.onCreate (BackgroundAudioService.java:474)
  at android.app.ActivityThread.handleCreateService (ActivityThread.java:3956)
  at android.app.ActivityThread.-wrap5 (Unknown Source)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2109)
  at android.os.Handler.dispatchMessage (Handler.java:109)
  at android.os.Looper.loop (Looper.java:166)
  at android.app.ActivityThread.main (ActivityThread.java:7555)
  at java.lang.reflect.Method.invoke (Method.java)
[...]

PR #345, opened 2020-05-07
Potential fix for #340

bsutton:
Looks like #340 is caused by a user clicking a media button before Flauto is fully initialised. 
This is likely to be prevalent on phones that map hard buttons to media events (volume/pause/resume/play).

We now discard any media button events until Flauto is fully initialised.

This pr also has fixes for the slot allocation not handling lazy initialisation in the dart code.

NOTE; this may still be an issue in the IOS code.
@Larpoux  can you take a look at the changes I've made to the slot initialisation in the SoundPlayer/SoundRecorder and TrackPlayer and replicate those changes to iOS.

Includes a number of bug fixes including for the demos.

Removed onFinished in favour of onStopped. It turns out you can't actually tell the difference as the plugin calls onFinished whether the player stops naturally or is stopped via user or api. As such the onStopped is a better verb as onFinished is mis-leading (it implies that the recording finished to completion). This will also simplify cleanup as users no longer have to hook both onStopped and onFinished.

Fixed an IllegalStateException when the player completes due to a race condition.

Upgraded to flutter 1.17 including required package changes.

### DIFF
--- a/android/src/main/java/com/dooboolab/fluttersound/BackgroundAudioService.java
+++ b/android/src/main/java/com/dooboolab/fluttersound/BackgroundAudioService.java
@@ -66,6 +66,8 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.List;
 import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+
 
 public class BackgroundAudioService
 	extends MediaBrowserServiceCompat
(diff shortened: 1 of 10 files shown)


3mrdev:
I am waiting to test it

bsutton:
We are currently having conversations over how to proceed with the dev branch I've been working on.
We are have a little trouble reaching a consensus.

If you want to test it out you can do a fork/clone of my branch at bsutton/flutter_sound.
Have a look at the read.me has it has notes on migrating to the 5.0 branch (dev).

Brett

Larpoux:
We actually have a `little` management problem.
Until this issue is fixed, I think it is not correct for you, @bsutton , to do ads for your fork in Flutter Sound project.

bsutton:
There's fixes can probably be back ported and I've never been one to stand
in the way of a user testing code.

Larpoux:
You told us that you consider V4 as a LTS (Long term support) version.
A crash is something completely unacceptable and we are waiting for a push into V4 from you.

Suggest to a user to use your experimental fork is irresponsible.

bsutton:
Irresponsible?

To suggest they use it yes, to suggest they test it no.

As to a pr for the code changes I had assumed that you would do it but I
can arrange it if you want.

Larpoux:
Yes I do want.

bsutton:
I will try to organise it over the weekend.

bsutton:
Superseded by #357

PR #348, opened 2020-05-08
Fixed: parameters androidOutputFormat and androidEncoder are ignored

greymag:
Now parameters `androidOutputFormat` and `androidEncoder` used as is if defined, and automatically resolved if `null` (this is default behaviour, so nothin changed by default).

Some additional information.
I need this parameters because of bug on some Android 5 devices - audio recorded but not played with default output format, which is `OutputFormat.AAC_ADTS`. It is working if use `OutputFormat.DEFAULT` or `OutputFormat.MPEG_4`, but I'm not sure that it is a wright way to fix it. That's why I decide to not change android recorder implementation, but just add an ability to set output format from application code (since this parameter defined in package).

### DIFF
--- a/android/src/main/java/com/dooboolab/fluttersound/FlutterSoundRecorder.java
+++ b/android/src/main/java/com/dooboolab/fluttersound/FlutterSoundRecorder.java
@@ -392,11 +392,11 @@ public class FlutterSoundRecorder
 			Integer      sampleRate          = call.argument ( "sampleRate" );
 			Integer      numChannels         = call.argument ( "numChannels" );
 			Integer      bitRate             = call.argument ( "bitRate" );
-			int          androidEncoder      = call.argument ( "androidEncoder" );
+			Integer      androidEncoder      = call.argument ( "androidEncoder" );
 			int          _codec              = call.argument ( "codec" );
 			t_CODEC      codec               = t_CODEC.values ()[ _codec ];
 			int          androidAudioSource  = call.argument ( "androidAudioSource" );
-			int          androidOutputFormat = call.argument ( "androidOutputFormat" );
+			Integer      androidOutputFormat = call.argument ( "androidOutputFormat" );
 			final String path                = call.argument ( "path" );
 			_startRecorder ( numChannels, sampleRate, bitRate, codec, androidEncoder, androidAudioSource, androidOutputFormat, path, result );
 		}
@@ -405,7 +405,7 @@ public class FlutterSoundRecorder
 	}
 
 	public void _startRecorder (
-		Integer numChannels, Integer sampleRate, Integer bitRate, t_CODEC codec, int androidEncoder, int androidAudioSource, int androidOutputFormat, String path, final Result result
+		Integer numChannels, Integer sampleRate, Integer bitRate, t_CODEC codec, Integer androidEncoder, int androidAudioSource, Integer androidOutputFormat, String path, final Result result
 	                           )
 	{
 		final int v = Build.VERSION.SDK_INT;
@@ -434,8 +434,17 @@ public class FlutterSoundRecorder
 			}
 			mediaRecorder.reset();
 			mediaRecorder.setAudioSource ( androidAudioSource );
-			androidEncoder      = codecArray[ codec.ordinal () ];
-			androidOutputFormat = formatsArray[ codec.ordinal () ];
+
+			// If encoder is defined, then use it, otherwise use the auto detection by codec
+			if (androidEncoder == null) {
+				androidEncoder      = codecArray[ codec.ordinal () ];
+			}
+
+			// If output format is defined, then use it, otherwise use the auto detection by codec
+			if (androidOutputFormat == null) {
+				androidOutputFormat = formatsArray[codec.ordinal()];
+			}
+
 			mediaRecorder.setOutputFormat ( androidOutputFormat );
 
 			if ( path == null )
--- a/lib/flutter_sound_recorder.dart
+++ b/lib/flutter_sound_recorder.dart
@@ -284,15 +284,19 @@ class FlutterSoundRecorder {
     return fout.path;
   }
 
+  /// Starts record.
+  /// 
+  /// If [androidOutputFormat] is null (default) - it will be auto detected by [codec].
+  /// If [androidEncoder] is null (default) - it will be auto detected by [codec].
   Future<String> startRecorder({
     String uri,
     int sampleRate = 16000,
     int numChannels = 1,
     int bitRate = 16000,
     t_CODEC codec = t_CODEC.CODEC_AAC,
-    AndroidEncoder androidEncoder = AndroidEncoder.AAC,
+    AndroidEncoder androidEncoder,
     AndroidAudioSource androidAudioSource = AndroidAudioSource.MIC,
-    AndroidOutputFormat androidOutputFormat = AndroidOutputFormat.DEFAULT,
+    AndroidOutputFormat androidOutputFormat,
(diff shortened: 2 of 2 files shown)


bsutton:
I've just removed these values from the dev branch so perhaps we should take a moment and work out if they are needed.
The output format is being taken from the codec so in theory the correct fix is to change the codec which will then change the output format.
Did you try changing the codec to mpeg?

greymag:
Parameter `codec` in `startRecorder` is `t_CODEC` enum:
```dart
enum t_CODEC {
  DEFAULT,
  CODEC_AAC,
  CODEC_OPUS,
  CODEC_CAF_OPUS, // Apple encapsulates its bits in its own special envelope : .caf instead of a regular ogg/opus (.opus). This is completely stupid, this is Apple.
  CODEC_MP3,
  CODEC_VORBIS,
  CODEC_PCM,
}
```

`CODEC_MP3` - marked as "Not implemented". Which one should I try (on Android)?

Larpoux:
@greymag ,
Hi Grey 😸 ,

I am looking to your Pull Request. I am very late to do that. Excuse us.
We were really **very** busy with the V5 release.
I know, this is not an excuse... 👎 

Your patch seems good for Flutter Sound V4. I agree with your Pull Request. I am going to merge it with the V4 branch, if others developers are OK (@bsutton and @hyochan ).

For the V5, the specific Android parameters has been abandoned. Now, there is no more OS specific parameters in startRecorder().

I suggest to use correctly the codec value `defaultCodec`, which is a value not actually really used. If the caller specifies `defaultCodec` as a codec, we will format the sound with Android default format. For me this sounds correct. I can do that if others are OK (@bsutton and @hyochan ).

In fact we can do the same for the V4, instead of your PR. What do you think for this ?

[again, excuse us for our very late answer)

Larpoux:
@greymag : could you look to #362 (PR against V5) ?

greymag:
> I suggest to use correctly the codec value `defaultCodec`, which is a value not actually really used. If the caller specifies `defaultCodec` as a codec, we will format the sound with Android default format. For me this sounds correct. I can do that if others are OK (@bsutton and @hyochan ).
> 
> In fact we can do the same for the V4, instead of your PR. What do you think for this ?

It's still will not work for `CODEC_AAC` and parameters in dart code still remain fake, but it's give an ability to fix problem in application code, as well as mine. So, it's ok for me =)

Larpoux:
The problem is for V5 : `androidOutputFormat` parameter does not exist any more, and I would not like to reintroduce it because it is OS dependant.

For V4, your patch should work.

greymag:
I understand. Here I am talking only about v4 version.

bsutton:
If this is a problem with api versions post 10 then I'm concerned that we
don't actually understand the problem do the fix is likely to be incorrect.

Fyi we have anymore information in the problem.

Do you see any errors when it fails?

If you call the method that checks if a codec is supported guess it returns
true.

I don't like putting fixes in if we don't actually understand the problem.

Larpoux:
Rejected by @bsutton .
We are now waiting for his own PRs (for V4 and V5)

greymag:
> If this is a problem with api versions post 10 then I'm concerned that we don't actually understand the problem do the fix is likely to be incorrect. 
> Fyi we have anymore information in the problem. 
> Do you see any errors when it fails? 

Record didn't fails at all. It starts and complete without any error and create record file. But I can't play this record on any device (neither Android and iOS). And not only with this plugin - default player can't play it too.

However on the desktop I will able to play it, but not on any of tested mobile phones. 

> If you call the method that checks if a codec is supported guess it returns true.

Yes

Here the log when I try to play record:
```
05-24 23:01:53.291 8048-8369/? I/AudioPlayer: start of Playback, useOffload 0
05-24 23:01:53.291 8048-30041/? W/SoftAAC2: aacDecoder_ConfigRaw decoderErr = 0x2003
05-24 23:01:53.292 8048-30042/? E/OMXCodec: [OMX.google.aac.decoder] OMX_EventError(0x80001001, 8195)
05-24 23:01:53.292 8048-8369/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.292 8048-8369/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.292 8048-8369/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.292 8048-8369/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.292 8048-8369/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.292 8048-8369/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.292 8048-8369/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.293 8048-8369/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.295 29316-29466/ru.innim.notes I/flutter: startPlayer result: /data/data/ru.innim.notes/app_flutter/sound_records/sound_record_1590350507295.aac
05-24 23:01:53.393 8048-30043/? E/AudioSink: received unknown event type: 1 inside CallbackWrapper !
05-24 23:01:53.400 8048-30043/? E/AudioSink: received unknown event type: 1 inside CallbackWrapper !
05-24 23:01:53.456 29316-29333/ru.innim.notes E/MediaPlayer: error (1, -2147483648)
05-24 23:01:53.456 8048-30040/? I/AudioPlayer: Pause Playback at 0
05-24 23:01:53.458 29316-29316/ru.innim.notes E/MediaPlayer: Error (1,-2147483648)
05-24 23:01:53.459 29316-29316/ru.innim.notes E/MediaPlayer: Attempt to call getDuration without a valid mediaplayer
05-24 23:01:53.459 29316-29316/ru.innim.notes E/MediaPlayer: error (-38, 0)
05-24 23:01:53.460 8048-8048/? I/AudioPlayer: reset: mPlaying=0 mReachedEOS=1 useOffload=0
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
05-24 23:01:53.567 8048-8352/? E/AudioPolicyManager: unknown stream type
```

bsutton:
Looks like this is actually a bug in android 5.

A couple of references that seem relevant:

https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/-/issues/189
https://github.com/google/ExoPlayer/issues/178
https://lists.freedesktop.org/archives/gstreamer-bugs/2017-February/194651.html
http://gstreamer-devel.966125.n4.nabble.com/Playing-ts-files-on-android-with-playbin-td4676878.html
https://github.com/mstorsjo/fdk-aac/pull/73

Larpoux:
Fixed on V5.1.0

PR #350, opened 2020-05-08
Fixed: iOS SoundPlayer implementation did not return error when fails.

greymag:


bsutton:
I'm not an iOS expert but broadly looks ok to me.

Can you apply a patch to 5.0.

PR #362, opened 2020-05-23
Fix #348

Larpoux:
This PR is supposed to fix @greymag PR #348.

@greymag
@bsutton 
@hyochan 

This is my first Pull Request agains the V5 😃 
Is it OK for you ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 5.1.0
+
+- On iOS : the "NowPlaying" info on the lockscreen is removed when the sound is fnished or when the App does a ```stopPlayer()```
+- On iOS : the "NowPlaying" progress bar on the lockscreen is uptodated when the App does a ```seekToPlayer()``` [#364](https://github.com/dooboolab/flutter_sound/issues/364)
+
 ## 5.0.1
 
 - Flutter Sound V5 is published under the LGPL license.
--- a/README.md
+++ b/README.md
@@ -97,14 +97,18 @@ When you fill an issue, we try to answer something in less than 48h. Of course, 
 
 ## TODO
 
-- [X] Record PCM on Android
-- [X] Rewrite the documentation
-- [ ] Small examples, smaller than the actual demox.app
+- [ ] Record and playback raw PCM both on Android and iOS
 - [ ] Record OPUS on Android
+- [ ] Record MP3 both on Android and iOS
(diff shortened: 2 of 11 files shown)


bsutton:
Can I suggest a slightly different approach.

The problem is essentially that the AAC coded was not introduced until api level 10.

If I understand correctly, selecting the DEFAULT codec results in the audio be encoded to AAC.

I'm not a big fan of the idea of a 'default' codec as its not clear what codec you will get back once recording is completed.

In the 5.0 branch I've removed any reference to a 'default' codec.

To fix this particular issue I would suggest that if the dart code passes down the AAC codec then we switch to the 'DEFAULT' codec in the java code.
We probably should put in an api version check (i.e. only do the switch if api is less than 10) just in case the meaning of 'DEFAULT' has changed over time.

This means that a user always explicitly chooses a codec, so they know what codec is being recorded, but we also support devices using < api 10.

Thoughts?

Larpoux:
Yes, @bsutton .
Your suggestion will fix a bug when the user specify AAC, but AAC is not supported by the OS.
But I would like also that DefaultCodec will be really the OS default, and not something that we arbitrary force.

Can we do both ?

greymag:
> The problem is essentially that the AAC coded was not introduced until api level 10.

The problem in my PR is reproduced on Android 5 (API 21).

greymag:
About this PR.

Problem appears with default parameter `codec:  t_CODEC.CODEC_AAC`.

So, in current (v4) version it lead's to
```
codec = MediaRecorder.AudioEncoder.AAC
outputFormat = MediaRecorder.OutputFormat.AAC_ADTS
```
and this set of parameters doesn't work correctly.

If I change output to `MediaRecorder.OutputFormat.DEFAULT` - it works.

Changes in this PR probably fix problem with default `code` parameter (I can check it later on the problem device), but codec `Codec.aacADTS` still will lead to corrupted record.

Larpoux:
Would it be OK for you to use t_CODEC.DEFAULT
and FS will use the Android default both for the Encoder and the FileFormat ?

bsutton:
The problem with allowing a default is that there is no way to discover
what format the recording occurred in.

5.0 currently requires a codec. I would prefer to leave it that way and
find out what is actually causing this problem. That information may
indicate other changes are required.

Larpoux:
Rejected by @bsutton .
We are now waiting for his own PR.

Issue #374, opened 2020-05-30
[BUG]  When the App does a "StopPlayer()" DURING a "startPlayer()"


Larpoux:
## Flutter Sound Version : 

- Version number ?
V4 & V5

- **FULL** or **LITE** flavor ?
FULL

----------------------------------------------------------

## Severity

- Result is not what expected 

--------------------------------------------------------

## Platforms you faced the error 

- iOS 
- Real device 

----------------------------------------------

We have a very serious bug when the App does a "StopPlayer()"  **DURING** a "startPlayer()" is currently in progress.
This occurs easily if startPlayer() does ```await``` on FFmpeg before calling the real startPlayer() OS code.

Larpoux:
Fixed in 5.1.0

PR #375, opened 2020-05-31
V4 #374 stop player duging start player

Larpoux:
Patch to fix a bug when the App does a ```stopPlayer()``` during a ```startPlayer()```.
Bug #374


Larpoux:
V 5.1.0

PR #442, opened 2020-08-28
Future V6.0

Larpoux:
Future V6.0


android/build.gradle
```diff
@@ -28,7 +28,7 @@ android {
     compileSdkVersion 28
 
     defaultConfig {
-        minSdkVersion 24 // 16
+        minSdkVersion  21
```
Alezhka, line 31: ❤️

Larpoux: Thank you @Alezhka for your Review on the future Flutter Sound V6.0. 👍 

I hope that V6.0 will be a great major FS version. 😄 
It will have support to **Record to a Flutter Stream** on ios and Android, and **playback from a Flutter Stream** on Android which is a feature that many, many, many Flutter Sound users are waiting for.

I will try also to support Flutter Sound on the Web Apps and Macos Desktop. But the work is actually not done.

Take care.

Alezhka: Good luck!

We are waiting for the release ☺️

Alezhka:
Congratulations! 🥰

PR #452, opened 2020-09-10
Correct copy/paste albumArtFile assignment error

gallafent:
Was previously being assigned from albumArtAssetString, clearly that stanza copied and incompletely edited into this one!

Spotted by Xcode ;)


Larpoux:
Thank you for your contribution, @gallafent

PR #471, opened 2020-09-29
update recase version

CRJFisher:
recase version is preventing us updating other libraries.

Thank you for this library, we will try to make more contributions in future


Larpoux:
Thank you very much for your Pull Request 👍

PR #529, opened 2020-11-28
Update logger package to ^0.9.4

apgapg:
Signed-off-by: Ayush P Gupta <ayushpguptaapg@gmail.com>

Currently i am using logger package in one of library in main project but due to old logger version in flutter_sound, it cause error of mismatch version


Larpoux:
Thank you @apgapg ,

I am very glad when someone does not only report a problem, but also post a fix. (This is not frequent, unfortunately).
Thank you very much for your contribution.
Your fix will be include in version V6.4.3. I probably will release this version next week.

Issue #642, opened 2021-03-09
[BUG]: IOS can't register flutterSoundPlayerManager or flutterSoundRecorderManager


jordygarcias:
## Flutter Sound Version : ^7.7.0+1

- **FULL** or **LITE** flavor ? FULL

- **Important**: Result of the command : ```flutter pub deps | grep flutter_sound```

![image](https://user-images.githubusercontent.com/58491711/110398470-94085f80-8039-11eb-99bc-6f587dfd197b.png)

----------------------------------------------------------

## Severity

Crash ? It crashes every time I try to initialize player or recorder. In the console is being triggered the follow messages:
- ERROR during registerWithRegistrar: flutterSoundPlayerManager != nil
- ERROR during registerWithRegistrar: flutterSoundRecorderManager != nil

--------------------------------------------------------

## Platforms you faced the error 

- iOS ? It happens on iOS Emulator

------------------------------------------------

**Describe the bug**
It is failing at registering the plugin for ios, it gets the follow messages in console:
ERROR during registerWithRegistrar: flutterSoundPlayerManager != nil
ERROR during registerWithRegistrar: flutterSoundRecorderManager != nil

The app builds and runs well, but when I try to initialize the player or the recorder, it crashes.
----------------------------------------------

# Logs!!!!
(**This is very important**. Most of the time we cannot do anything if we do not have information on your bug)
At the beginning it throws the follow messages:
- ERROR during registerWithRegistrar: flutterSoundPlayerManager != nil
- ERROR during registerWithRegistrar: flutterSoundRecorderManager != nil

When I try to initialize the recorder or player it throws:
- flutter: FS:---> openAudioSession
- flutter: ---> openAudioSession
- flutter: Resetting flutter_sound Recorder Plugin
- @resetPlugin
- iOS: ---> resetPlugin
- iOS: <--- resetPlugin

Then It crashes
-----------------------------------------------------------------

aavgeen:
I am getting the same logs:

flutter: FS:---> openAudioSession
flutter: ---> openAudioSession
flutter: Resetting flutter_sound Recorder Plugin
@resetPlugin
iOS: ---> resetPlugin
iOS: <--- resetPlugin

As soon as it hits the line 

```
await _mRecorder.openAudioSession();
```

I followed the simple_recorder.dart example in ^7.7.0.

ayush221b:
I am facing the same error in our project, any help @jordygarcias

touficzayed:
[...]

PR #686, opened 2021-05-22
Fix crash #642

touficzayed:
This is a workaround to fix crash #642.

I just added a return statement before we reach the code that crashes.

So far it seems that my app doesn't crash when I use it along just_audio.

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -23,6 +23,10 @@ toc: false
 
 - Null Safety
 
+## 7.8.0
+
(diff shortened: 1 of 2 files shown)


Larpoux:
This patch is OK for me. I do not really understand how it is possible to reach this error, but your fix cannot be armful.
If you confirm that it solves your problems I suggest to integrate your Pull Request in next Flutter Sound version. Maybe tomorrow.

Larpoux:
@touficzayed ,

Your Pull Request has been integrated in Flutter Sound 8.1.4
👍

PR #702, opened 2021-06-17
Update dev docs to use URL instead of SSH for git clone command

mhstoller:
If someone doesn't have SSH keys configured with GitHub, the command would fail so prefer URL instead. Maybe the docs should include instructions to fork the project and clone that fork instead; open for discussion.


Larpoux:
Your doc modification is now integrated in the current documentation

PR #722, opened 2021-07-11
Update simple_recorder.dart

letmebelazy:
I used your package and it was great experience. I referred to this example, and I found it had a redundancy and its function do nothing when pressing the buttons. I'm a newbie so I might be wrong. But I think your package is really great, and I just want to contribute to this project if me finding is correct. Thank you guys for sharing this wonderful work.

### DIFF
--- a/flutter_sound/example/lib/simple_recorder/simple_recorder.dart
+++ b/flutter_sound/example/lib/simple_recorder/simple_recorder.dart
@@ -57,11 +57,8 @@ class _SimpleRecorderState extends State<SimpleRecorder> {
       });
     });
 
-    openTheRecorder().then((value) {
-      setState(() {
-        _mRecorderIsInited = true;
-      });
-    });
+    // You already gave true to _mRecorderIsInited variable. So it's no use giving true to it again.
+    openTheRecorder();
     super.initState();
   }
 
@@ -132,19 +129,19 @@ class _SimpleRecorderState extends State<SimpleRecorder> {
   }
 
 // ----------------------------- UI --------------------------------------------
-
-  _Fn? getRecorderFn() {
+  // If this function returns a function, it will never do anything. We need to call next function instead of returning it.
+  void getRecorderFn() {
     if (!_mRecorderIsInited || !_mPlayer!.isStopped) {
-      return null;
+      return;
     }
-    return _mRecorder!.isStopped ? record : stopRecorder;
+     _mRecorder!.isStopped ? record() : stopRecorder();
   }
 
-  _Fn? getPlaybackFn() {
+  void getPlaybackFn() {
     if (!_mPlayerIsInited || !_mplaybackReady || !_mRecorder!.isStopped) {
-      return null;
+      return;
     }
-    return _mPlayer!.isStopped ? play : stopPlayer;
+    _mPlayer!.isStopped ? play() : stopPlayer();
   }
 
   @override


Larpoux:
Thank you very much @letmebelazy for your Pull Request.
I am very glad when someone wants to contribute to the Tau Development.
Unfortunately not many people wants to contribute and I feel very lonely.

I am looking to your pull request now.

Larpoux (commented):
I think that you missed the point : 
`getRecordFn()` and `getPlayerFn()` are not functions that are called when the user push the buttons.
Instead, they are functions which returns the functions that must be called when we push the buttons.
The functions are called each time there is a build() of the UI.

Larpoux:
Please consider those 3 snippets : 
```dart
onPressed: getRecorderFn(),
```
```dart
onPressed: getRecorderFn,
```
```dart
onPressed: (){getRecordFn()},
```
If you can master those tree examples, you will master a large part of Dart.

Good luck for your investissement on coding.

letmebelazy:
OMG... I had no idea about the difference. I must admit I bothered you. I'm really sorry. And thank you for informing me about a significant part of Dart.

letmebelazy:
The first one is a function call, so it executes immediately, and assigns its return value to the property 'onPressed'.
The second one is a function reference, and it is called when we click the button like a callback function.
The last one is a anonymous function, so when we click the button, the function is called. Consequently, it acts as the second one.

BTW, I wonder what you mean by the comment in the simple_record example. You wrote "It writes all the recorded data from a Stream to a File, which is completely stupid: if an App wants to record something to a File, it must not use Streams.". Then, what should I do instead of using Stream?

Larpoux:
> Consequently, it acts as the second one.

YES!!! It is exactly that. Your answer is perfect. Some developers (not you) code `onPressed: (){getRecordFn();},` (which is completely unreadable) instead of a simple `onPressed: getRecorderFn,`.



> Why is it bad? And, what should I do instead of using Stream?

I just wanted to say with my poor english, that if someone wants to record the microphone to a file it is better to just do a `startRecorder() toFile: ...` than a `startRecorder() toStream: ...` and then write the stream to a file.
This is more complicated and less powerful.



> I must admit I bothered you. I'm really sorry.

**NO!** .  Don't be sorry. You did not bother me. I like very much when a Flutter Sound user does not act as a consumer but tries to contribute and improve the library.
Unfortunately, contributions from other developers are very unfrequent.

PLEASE! Do not hesitate to contribute another time to the Flutter Sound. I really want other people on this project. There is so many things to do for Flutter Sound being the best Flutter library to do sounds.

letmebelazy:
I made an example using flutter sound. You can check it in my repository https://github.com/letmebelazy/voice_recorder. I just want you to know how many people are being benefited by your awesome work. Thank you!

Larpoux:
Thank you @letmebelazy to show me your App. This is really fine.
Just two remarks : 
- '선택한 파일을 삭제하시겠습니까?' is not very understandable for most of people in the word. If you do not want to support several languages (this is not very easy to do), probably better to do "Do you want to delete the selected file ?" in english.
- When I launch your App the second time, the Record button is greyed. I must go to the File list and return to the home screen to ungrey it.


Can I suggest a little improvement in your App ?  Use a Widget in the list that can do a "swipe to delete" to the left or to the right.
Also I am not really sure that it is OK to be able to delete the file during it is recorded.

PR #728, opened 2021-07-17
Check for valid file extension based on audio codec

mhstoller:
The only problem is when using `Codec.defaultCodec` I'm not sure how to determine what codec the system decided to use. If you choose 'Codec.mp3' you get a `_CodecNotSupportedException` exception, but if you use the default codec with .mp3 file extension, it works fine.

I need a way to get the codec being used by the OS before calling the new `isValidFileExtension` method so we can handle cases that use `Codec.defaultCodec`. I don't know how to do that, any advice is welcome

### DIFF
--- a/flutter_sound/lib/public/flutter_sound_recorder.dart
+++ b/flutter_sound/lib/public/flutter_sound_recorder.dart
@@ -506,6 +506,75 @@ class FlutterSoundRecorder implements FlutterSoundRecorderCallback {
     return result;
   }
 
+  ///Determines if the given file extension is valid for the audio codec.
+  ///This was introduced to mitigate issues with users trying to record to MP3
+  bool isValidFileExtension(Codec codec, String? extension) {
+    if (extension != null) {
+      var _validExtensions;
+      switch (codec) {
+        case Codec.aacADTS:
+          _validExtensions = ['.aac', '.adt', '.adts'];
+          break;
+        case Codec.aacMP4:
+          _validExtensions = ['.aac', '.mp4'];
+          break;
+        case Codec.amrNB:
+        case Codec.amrWB:
+          _validExtensions = ['.amr', '.3ga'];
+          break;
+        case Codec.defaultCodec:
+          //assume the correct extension was used
+          return true;
+        case Codec.flac:
+          _validExtensions = '.flac';
+          break;
+        case Codec.mp3:
+          _validExtensions = '.mp3';
+          break;
+        case Codec.opusCAF:
+          _validExtensions = '.caf';
+          break;
+        case Codec.opusOGG:
+          _validExtensions = '.ogg';
+          break;
+        case Codec.opusWebM:
+          _validExtensions = '.webm';
+          break;
+        case Codec.pcm16:
+          _validExtensions = ['.aiff', '.caf', '.wav'];
+          break;
+        case Codec.pcm16AIFF:
+          _validExtensions = '.aiff';
+          break;
+        case Codec.pcm16CAF:
+          _validExtensions = '.caf';
+          break;
+        case Codec.pcm16WAV:
+          _validExtensions = '.wav';
+          break;
+        case Codec.pcm8:
+        case Codec.pcmFloat32:
+          _validExtensions = ['.aiff', '.caf', '.wav'];
+          break;
+        case Codec.pcmWebM:
+          _validExtensions = '.webm';
+          break;
+        case Codec.vorbisOGG:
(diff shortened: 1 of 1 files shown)


Larpoux (commented):
Thank you @mhstoller for your Pull Request.

I appreciate that you contribute to Flutter Sound maintenance.
I have two remarks for your PR : 
- I do not like an assert and I would prefer to throw an exception. When an Assert is triggered, it means there is a bug. But this is not a Flutter Sound bug. It happens sometimes that the App tries to record with a bad extension. Better to throw it an exception
- You do not check for "defaultCodec" . But this is really the case where there is problem. I do not expect many developer will write `startRecorder(codec: codec.opusOGG, toFile: foo.mp3);`. If he/she write something like that, he/she is really stupid. But the real problem is when the developer write `startRecorder(toFile: foo.mp3);` without specifying the Codec.

mhstoller:
This code by itself is not very useful, but if you or someone else can pass the actual codec (ex. mp3, opusOGG, etc) that gets used by `Codec.defaultCodec` into this function then you can throw a proper exception or handle it however you want.

Larpoux:
> This code by itself is not very useful

NO!!!!  I do not agree.
This code is really very useful. I did not realized that my control with "isCodecSupported" did not detect when the App uses the "DefaultCodec" and a bad extension.
I wrote in the documentation that the file extension must be consistent with the specified Codec, but this was not enough.
As you wrote in another thread, if two Flutter Sound users in a row are trying to record a file "foo.mp3", and receive an understandable Exception, it means that we must do something in our code.

This will be in next version (in a few days). Thank you.

davidAg9:
> Thank you @mhstoller for your Pull Request.
> 
> I appreciate that you contribute to Flutter Sound maintenance.
> I have two remarks for your PR :
> 
> * I do not like an assert and I would prefer to throw an exception. When an Assert is triggered, it means there is a bug. But this is not a Flutter Sound bug. It happens sometimes that the App tries to record with a bad extension. Better to throw it an exception
> * You do not check for "defaultCodec" . But this is really the case where there is problem. I do not expect many developer will write `startRecorder(codec: codec.opusOGG, toFile: foo.mp3);`. If he/she write something like that, he/she is really stupid. But the real problem is when the developer write `startRecorder(toFile: foo.mp3);` without specifying the Codec.

I see no reason to insult or pass your judgement to any developer , whether you are an expert or not ....unless you do not want this package to be used by the flutter community .Your aim to to provide a way for flutter developers of all kinds to be able to record audio .... I waste too much time of this as I see you are only a collaborator ....But if someone is using a product wrongly it simply means that either the documentation wasn't clear enough or easy to understand or your code isn't preventing such trivial mistakes by way of errors etc. I suggest you respect the views of everyone here ...as it will go a long way to help this package become a flutter favorite and a very usable one ....I like  although it has been constantly throwing unexpected errors in production mode and The flutter sound ui is not reponsive to say the least ....but I hope all these is in the mind of the original  owner of this repo.

Larpoux:
@davidAg9 ,

I do not understand your post. Nobody insult other people.
I try to do my best to maintain Flutter Sound and @mhstoller helped me.
We agree that there were a problem with `startRecorder()` : 
I did not realized that when the App developer use the `defaultCodec` and use an extension not supported (foo.mp3 for example), then the check "isValidEncoder()" was happy and later the App had an understandable Exception.
Last wee, we had two developers who tried to record MP3 and had an Exception which they do not understand.
This really means that something is bad in Flutter Sound and must be fixed.

This was bad, so @mhstoller  and me fixed that in the next release (in a few days).
I am not in a hurry to publish this new release, because nobody is actually blocked by this issue.

Note: doing a bad assertion is terrible. Not good. The App will crash without a good reason.
In the future version, I will throw a clear exception : 
- When the extension is not recognized ("foo.bla")
- When the extension is inconsistent with the codec specified (  codec: Codec.pcmWAV, toFile: 'foo.aac' )
- When the Codec is default and the extension is not correct  ( "foo.mp3" )

If you get other places where Flutter Sound has an understandable Exception, please either do a Pull Request (it would be great) or fill a new issue. But very often, the OS API gives error during startPlayer or startRecorder without being clear what was the reason. It is not always easy to investigate why the App has an Exception

Please, continue to use and better contribute to this library. I am sure that the future is bright for Flutter Sound.

Larpoux:
[Flutter Sound 8.3.4](https://tau.canardoux.xyz/changelog.html) is released.

PR #755, opened 2021-08-30
Update packages

aaassseee:
Update packages to latest version
removed package which is not in use


Larpoux:
Hi @aaassseee ,

Thank you for your Pull Request. You are right : flutter_spinkit is not used any more by Flutter Sound.
I am going to remove this unnecessary dependency.

Thank your for your contribution. We appreciate.

PR #763, opened 2021-09-08
Basic fixes to widget UI

jfkominsky:
Not a complete set of fixes, but corrects the most glaring issues with the widget UI. If you create a RecorderPlaybackController with both a recorder and a player, the recorder is now disabled when the player is playing, which heads off a bunch of potential error states. Similarly, when recording the player is disabled. 

The widget UI still has a number of incomplete features but the basic functions now work consistently. Still needs a proper maintainer but this should keep it functional for a bit.


jfkominsky:
Note: There is still some issue I cannot understand that arises when you exit the widget demo screen and then return to it. Something is being set null when it should not be. I will continue to investigate.

Issue #764, opened 2021-09-09
[BUG]: SoundRecorderUIState mysteriously becoming null


jfkominsky:
Filing this not because I expect anyone else to fix it but because I want to have a record of it so I'm sure I'm not hallucinating.

Basically there's a lot of failure states that arise in a RecordPlaybackController because it reads a SoundRecorderUIState object as null when it really shouldn't, and I have absolutely no idea how or why.

Quick overview: In order for a RecordPlaybackController to coordinate between a SoundRecorderUI and a SoundPlayerUI, it has the following internal variables:
```
  SoundRecorderUIState? _recorderState; 
  SoundPlayerUIState? _playerState;
```
These are assigned during the initialization of the SoundPlayerUI and SoundRecorderUI using the functions registerPlayer() and registerRecorder, which goes in two steps. First, there's an interface function that the UI widget actually calls:
```
void registerRecorder(BuildContext context, SoundRecorderUIState recorder) {
  RecorderPlaybackController.of(context)?._state.registerRecorder(recorder);
}
```
Then there's the function that is part of the RecordPlaybackControllerState _state that does the actual work (note: I am going to rename this stupid thing _registerRecorder because come on)
```
void registerRecorder(SoundRecorderUIState recorderState) {
    _recorderState = recorderState;

    // wire our local stream to take events from the recording.
    _recorderState!.dispositionStream!.listen(
        (recorderDisposition) => _localController.add(PlaybackDisposition(
            // TODO ? PlaybackDispositionState.recording,
            position: Duration.zero,
            duration: recorderDisposition.duration)));
  }
```
After this point, _recorderState should have a value, and there is no way I can see for it to be null.

The first time you access a page that contains this widget (e.g., in the example app), this all seems to work perfectly as intended, and you can access _recorderState however you want.

However, if you leave the page and then come back to it later, without restarting the app, something weird happens. registerRecorder is still called, and a non-null state is registered. I've managed to establish that using logging. Then we come to the stuff I added, which makes the player and recorder accessible or not depending on whether the other one is active, and something utterly bizarre happens. Here's the code I'm currently debugging with:

As part of RecordPlaybackControllerState :
```
void _onRecorderNew() {
    /// For the specific case of there already being something recorded and
    /// recording again, the play button needs to be removed.
    if (_playerState != null) {
      _playerState!.playbackEnabled(enabled: false);
    }
[...]

PR #765, opened 2021-09-10
Fixes and cleanups for widget UI

jfkominsky:
Fixed issue #764 in two ways:

1. Figured out that the mysterious null errors were likely arising due to some oddity of how InheritedWidget handles of(context), and put in a check that, under normal use cases, will solve the problem by manually updating the state that was previously returning null.
2. In the case that this fix is unsuccessful for whatever reason, made the null-checks gentler so they don't throw exceptions, and guarded against an edge case that could arise with the player by switching an assertion for an if/else. 

If used in the simple, intended way (a recorder and a player) the whole system should now just work and not throw exceptions. There are probably more complex cases where these fixes won't work, but if people run into them I expect they'll file BRs. I want to do more work on this widget eventually anyways so I'll deal with those issues then, but for my purposes at least this is working well enough.


Larpoux:
> 2\. and guarded against an edge case that could arise with the player by switching an assertion for an if/else.

When I started to work on this project, I very often coded "assert()", because I thought that the condition should never arrive if the App is correctly coded. It was a big mistake : the Appp and the Plugin are coded by two different groups, and I cannot master what the App does. When Assert fails, the App developer consider that it is a bug in flutter sound. Now I do not code Assert() when I cannot be sure that the condition never arrives, and I throw instead an exception.

> I want to do more work on this widget eventually anyways

Thank you @jfkominsky  for your contribution. We all appreciate.

Larpoux:
Flutter Sound 8.3.12 is released with your Pull Request.  Thank you again.

PR #821, opened 2022-01-18
Added read decibel value from recorder on web

GeorgeAmgad:
The values that are set to the analyser still needs to be checked.
I tried to set it so its behaviour could be as close as possible to that of android.


Larpoux:
Thank you George for your contribution.
We really appreciate when a developer contribute to Flutter Sound

Larpoux:
Flutter_sound 8.5.0 is released with your PR merged inside

aleslanik2:
Hi, i think in current stable is not PR merged. When i look to my_project/build/web/assets/packages/flutter_sound_web/src/flutter_sound_recorder.js so on line 413 is that block of code 

```
function () {
         var now = new Date().getTime();
         var distance = now - me.countDownDate;
         me.callbackTable[CB_updateRecorderProgress](me.callback, me.deltaTime + distance, 0);
},
```

PR #870, opened 2022-04-01
VoiceProcessing feature for iOS

acalatrava:
[Linked](https://github.com/Canardoux/flutter_sound_core/pull/4)

While developing a VoIP app I noticed that the audio was too low. After some research I see that activating the VoiceProcessingIO AudioUnit will improve the speech audio a lot, so I added the function to this lib.

I'm not sure if this is possible on Android but I will investigate it as well soon.

### DIFF
--- a/flutter_sound/ios/Classes/FlutterSoundPlayer.mm
+++ b/flutter_sound/ios/Classes/FlutterSoundPlayer.mm
@@ -123,6 +123,8 @@
 - (FlutterSoundPlayer*)init: (FlutterMethodCall*)call
 {
         flautoPlayer = [ [FlautoPlayer alloc] init: self];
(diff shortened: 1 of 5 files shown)


Larpoux:
Hi Antonio,

Thank you for your Pull Request.
Can you explain very briefly what is the purpose of the flag *voiceProcessing* ?
I looked on internet, but I found that Apple documentation is almost un-existing.

Also I am not very glad to implement something only for ios.
And if we really need this new flag, we must modify the API documentation.

Larpoux:
@acalatrava 

Antonio, Flutter Sound 9.2.2 is released with your Pull Request integrated.
Thank you again for your contribution/

PR #904, opened 2022-06-03
iOS Voice processing enabled setting bug fix

vedranivicatos:
Voice processing enabled setting bug fix: changed from bool cast that would **always evaluate to true/YES** no matter what was received from dart side, to NSString description compare that works properly now. This wrong mapping was causing the voice processing to always be `true` and this in turn caused crashes when using multiple instances of Flutter Sound Player (issue I've opened: [bug](https://github.com/Canardoux/flutter_sound/issues/900#issue-1248767871 )).

### DIFF
--- a/flutter_sound/ios/Classes/FlutterSoundPlayer.mm
+++ b/flutter_sound/ios/Classes/FlutterSoundPlayer.mm
@@ -124,7 +124,7 @@
 {
         flautoPlayer = [ [FlautoPlayer alloc] init: self];
         flutterSoundPlayerManager = pm;
-        bool voiceProcessing = (bool)call.arguments[@"voiceProcessing"];
+        bool voiceProcessing = [[call.arguments[@"voiceProcessing"] description] isEqualToString: @"1"];
         [flautoPlayer setVoiceProcessing: voiceProcessing];
         return [super init: call]; // Init Session
 }


Larpoux:
Thank you @vedranivicatos for your Pull Request.
We appreciate your bug fix.
I am always impressed when someone is able to debug Flutter_sound_core without any help from me.
I will look to your patch and merge it into a new version tomorrow.

vedranivicatos:
@Larpoux You guys are welcome. I'm more than happy to contribute to such a project. However, this is just a necessary small fix. There is a bigger [issue](https://github.com/Canardoux/flutter_sound/issues/900#issue-1248767871) that this was causing that I opened a couple of days ago. This needs to be addressed as well.

KwonsooMoon:
Hi, thank you for your debugging and fix, I also waiting for this bug fixed.
But, it doesn't look safe  to compare description with hardcoded string.
How about casting to NSNumber and using boolValue? 
Like,
> bool voiceProcessing = ((NSNumber *)call.arguments[@"voiceProcessing"]).boolValue;

Official flutter [doc](https://docs.flutter.dev/development/platform-integration/platform-channels?tab=type-mappings-obj-c-tab) says flutter bool will be changed into NSNumber.

Larpoux:
@vedranivicatos : I added the boolean parameter "enableVoiceProcessing" inside startRecorder().
I think that this parameter is only when recording from mic and not relevant for the Player.

Please reopen the PR if you do not agree.

PR #908, opened 2022-06-16
fix: correct log level key

wurikiji:
A log level is written to `'level'` not `'logLevel'`.


Larpoux:
Thank you @wurikiji for your contribution.
Your patch is OK: the key "logLevel" was incorrect.
Your PR is now merged inside Flutter Sound.
I found another bug when logging from Android or iOS : the Logger Level Enum was renumbered int the last Logger version.
This issue will be fixed in Flutter Sound 9.4.7

PR #944, opened 2023-01-11
fix for VoiceProcessing feature

acalatrava:
This is a simple fix to the VoiceProcessing features since the `bool` variable was not passed correctly I switched it to a `int` value.


acalatrava:
It seems it was tried to be fixed on #904 but I think this is a better way to do it

Larpoux:
I think that the code is now correct inside Flutter Sound 9.4.5.
Please reopen your PR if you do not agree with the current code.

PR #945, opened 2023-01-12
Implemented setSpeed for startPlayerFromStream

acalatrava:
Since I'm using this for a VoIP app I needed to speed up the playback rate of the stream when there is a delay between the two parties. Because of this I implemented it both for Android and iOS and increase the blockSize to 8192 on Android to allow up to x3 speed on the stream.


Larpoux:
This PR is merged inside Flutter Sound 9.4.5.
Thank your @acalatrava for your contribution.
This is a great PR : I am impressed how you was able to update flutter Sound Core. This is not an easy task because FlutterSoundCore is not managed under pub.dev.

Thank you again. Your PR works very good without any modifications.

PR #949, opened 2023-02-21
fix:Flauto.androidActivity caused memory leak

caodingpeng:
Flauto.androidActivity is  a static variable. It has a strong reference to the connected activity.
when the activity destoryed, memoy leaks.


Larpoux:
@caodingpeng  : thank you for your contribution.
Unfortunately, I do not master correctly Android and I am not able to understand this code.

I merge your patch into the next Flutter Sound version (9.3.9), and I just hope that it is OK, because I really cannot validate it.

Thank you again for your contribution

PR #974, opened 2023-06-20
Update README.md

royesc:


Larpoux:
I removed most of the Ukrainian pictures in the lat versions

PR #980, opened 2023-07-02
fix recorder is not logging native debug messages

HaoCherHong:
Fix that messages from `FlautoRecorder.mm` in the [flutter_sound_core](https://github.com/Canardoux/flutter_sound_core) is not logging to console


Larpoux:
@HaoCherHong : thank you for your contribution.
I am OK with your patch. It will be merged into the next FlutterSound version (9.3.9)

PR #984, opened 2023-08-16
[Feature] macOS support

efraespada:
I made some minor changes in `flutter_sound_core` to make it work, so I included in the `macos`  folder the core library with these modifications.

For a proper test I needed to implement the `permission_helper` for Apple:

```yml
  permission_handler:                           # android   ios                         windows
    git:
      url: https://github.com/riccardo-lomazzi/flutter-permission-handler.git
      path: permission_handler
  permission_handler_web:                       #                                 web
    git:
      url: https://github.com/johnwcchau/permission_handler_web.git
  permission_handler_apple:                     #           ios           macos
    git:
      url: https://github.com/riccardo-lomazzi/flutter-permission-handler.git
      path: permission_handler_apple
```


Larpoux:
@efraespada ,
Thank you for your contribution. I am really fan of a support to MacOS. It would be very great.
But I do not think it is a good idea to duplicate code from FlutterSound core. The maintenance would be a nightmare.

Do you think it is possible to support both iOS and MacOS with the same core Code ?

efraespada:
Sure, it was a minimal change for make it work in macOS. If I remember well, I had to remove a non-used enum.

Larpoux:
@efraespada  : it would be great if someone does something to support MacOS. Unfortunately I cannot merge your PR because it is not possible to duplicate a such important code from FlautterSoundCore.

Please re-submit a new PR if you can do something for that.
Thank you again

PR #993, opened 2023-11-16
Update FlutterSoundManager.java

forzaZhuang:
111


Larpoux:
Hi @forzaZhuang ,

But I am not sure to agree with your patch:
you want to silently trap all exceptions. But we clearly understand why such an exception and at least do an error diagnose. 
I am going to keep this PR open, but I really understand an answer to this post.

Larpoux:
@forzaZhuang : it would be great if someone does something to support MacOS. Unfortunately I cannot merge your PR because it is not possible to duplicate a such important code from FlautterSoundCore.

Please re-submit a new PR if you can do something for that.
Thank you again

PR #1000, opened 2023-12-14
Fixes an issue in FlutterSoundPlayer -init:playerManager:  where the …

timcschmidt:
…voiceProcessing argument is always evaluated as true.

### DIFF
--- a/flutter_sound/ios/Classes/FlutterSoundPlayer.mm
+++ b/flutter_sound/ios/Classes/FlutterSoundPlayer.mm
@@ -124,7 +124,7 @@
 {
         flautoPlayer = [ [FlautoPlayer alloc] init: self];
         flutterSoundPlayerManager = pm;
-        bool voiceProcessing = (bool)call.arguments[@"voiceProcessing"];
+        bool voiceProcessing = ((NSNumber *)call.arguments[@"voiceProcessing"]).boolValue;
         [flautoPlayer setVoiceProcessing: voiceProcessing];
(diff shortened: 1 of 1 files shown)


timcschmidt:
This prevents mere playback requesting microphone access on ios.

Larpoux:
Thank you @timcschmidt for your contribution. I really appreciate.
There is clearly a bug there.

My problem is now that when I call 
```
if (![outputNode setVoiceProcessingEnabled:YES error:&err])
```
then, later, when I call 
```
[engine connect: playerNode to: outputNode format: outputFormat];
```
I get an exception "required condition is false: IsFormatSampleRateAndChannelCountValid(format)"

I am not sure that this is tied with your patch. But I do not really understand the iOS library.
If you have an idea of what is happening I would be grateful.

Larpoux:
in another test 
when I call : 
```
     AVAudioInputNode* inputNode = [engine inputNode];
     AVAudioFormat* inputFormat = [inputNode outputFormatForBus: 0];
     double sRate = [inputFormat sampleRate];
```
I get "sRate is 0" which is not correct.

Larpoux:
My feeling is that those two errors are not tied to your patch.
But I do not thing iOS library is very understandable . I am lost.

Larpoux:
Thank you again @timcschmidt for your contribution.
Your patch seems correct and it is now merged into Flutter Sound 9.3.8,
So I am closing your PR.
If you can look to #1019 I would be grateful.

PR #1031, opened 2024-05-02
Fix volume level on Web (#838, #862)

### LINKED ISSUE #838: [BUG]: onProgress not returning decibels for Web
## Flutter Sound Version : 

- FULL

- **Important**: Result of the command : ```flutter pub deps | grep flutter_sound```
|-- flutter_sound 9.1.2
|   |-- flutter_sound_platform_interface 9.1.2
|   |-- flutter_sound_web 9.1.2
|   |   |-- flutter_sound_platform_interface...
----------------------------------------------------------

## Severity
- Result is not what expected

--------------------------------------------------------

## Platforms you faced the error 
- Flutter Web

------------------------------------------------

**Describe the bug**
The onProgress callback is called but the decibels value is always 0 when running on web. Also https://flutter-sound.canardoux.xyz/web_example/#/ Recorder onProgress does not shows the decibels

**To Reproduce**
1. Go to https://flutter-sound.canardoux.xyz/web_example/#/
2. Click on Recorder onProgress
3. Click on Go
4. See error - no db value showing.
-----------------------------------------------------------------

### LINKED ISSUE #862: [BUG]: using on web, and flutter_sound_recorder.js is still using old version in Flutter project
## Flutter Sound Version : 

v 9.1.7

----------------------------------------------------------

## Severity

- App does not function without the volume being sent in the last parameter.

--------------------------------------------------------

## Platforms you faced the error 

- Flutter Web

------------------------------------------------

**Describe the bug**

using   flutter_sound: ^9.1.7 in pubspec.yaml, and yet the "flutter_sound_recorder.js" file that gets included is still the one that does not have the CB_updateRecorderProgress code fixed. i.e. it is still coming in as:

`me.callbackTable[CB_updateRecorderProgress](me.callback, me.deltaTime + distance, 0);
`

even though I can see if is fixed in the repo.

**To Reproduce**
Include the flutter_sound library on a flutter web app, and do anything that needs the volume. The parameter is hardcoded to 0.
----------------------------------------------

# Logs!!!!

No log needed... the code is correct in "flutter_sound_recorder.js" in the repo, but when flutter pub get is installing the library, the wrong .js file is coming down.

-----------------------------------------------------------------

bloemy7:
This fixes the volume level callback on Web. The file was correctly fixed and merged in the web folder in the `js` folder, but not in the `src` folder. This simply copies it over. I've tested and it solves the issue.

Resolves #838 
Resolves #862

### DIFF
--- a/flutter_sound_web/src/flutter_sound_recorder.js
+++ b/flutter_sound_web/src/flutter_sound_recorder.js
@@ -186,7 +186,7 @@ class FlutterSoundRecorder {
         }
 
 
-        async startRecorder(path, sampleRate, numChannels, bitRate, bufferSize, enableVoiceProcessing, codec, toStream, audioSource) {
+        async startRecorder(path, sampleRate, numChannels, bitRate, bufferSize, enableVoiceProcessing ,codec, toStream, audioSource) {
                 this.callbackTable[CB_recorder_log](this.callback, DBG, 'startRecorder');
                 //var constraints = { audio: true};
                 //var chunks ;//= [];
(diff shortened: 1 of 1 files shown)


bloemy7:
@Larpoux could this get merged in please? Thank you.

Larpoux:
Hi @bloemy7 👍 
Thank you very much for your contribution.
Your PR is now merged inside Flutter Sound 9.4.7

I agree that the JS scripts are not clean.
We have duplicates scripts at three (!!!) different places : 
- Flutter_sound_core/web/js/flutter_sound
- Flutter_sound_web/js/flutter_sound
- flutter_sound/example/web/flutter_sound

This is dirty.
I remember that I duplicated the scripts into flutter_sound/example: It was not easy to do symbolic links between flutter_sound_core and flutter_sound/examples because it is two different git repo. The problem is when I update the js scripts during an example run (during debug) chrome browser updates the example scripts and not the flutter_sound_core scripts.
I do not remember why it is also duplicated inside flutter_sound_web.

I am going to clean all that. At least we must give an error during bin/new-version.sh if the three directories are not synchronized. But I would prefer to remove those duplications.

PR #1053, opened 2024-07-02
Fix incorrect method names in the doc

iktakahiro:
Corrected the method names for open and close in the example using FlutterSoundRecorder.


Larpoux:
Thank you @iktakahiro for your time fixing the doc.
I really appreciate. Your update is now on line.

Note: the documentation is not in a good state, and someone should work on it.

PR #1078, opened 2024-08-22
Delete introduction to discontinued tau_sound package

tatsuyafujisaki:


Larpoux:
Thank you for your contribution. The doc is in bad shape and really needs to be cleaned up. I began to clean it up but can’t release it because it is tied with a future Flutter Sound 10.0.

Thank you again. Your contribution is very helpful.

